PR #6, opened 2020-05-08
adding ports-refining intergration files

ah1053:
1. New Dll 2.0.2
2. IsEdgeEnabled modified with UnityEditor
3. PlatformIntegration.cs modified for UnityEditor 
**Problem**
(using UnityEditor in iOS Mode or AndroidMode) >> Network interface defaults to MacNetInterface >> MacInterface have Cellular "en0" ,so IsEdgeEnabled fail to recognize its not EdgeCompatible.
**Solution**
 if Any OS that is not EdgeCompatible set NetInterface to empty NetInterface
4. Adding Port Choices to Editor Script and MobiledgeX Settings and optional argument for GetWebSocketConnection.
5. Adding "MobiledgeX :" to Error logs for easier debugging experience from developers prespective.


MobiledgeX/Editor/MobiledgeXEditorWindow.cs
```diff
@@ -350,7 +354,7 @@ void AddMobiledgeXPlugins()
                   }
                   catch (Exception e)
                   {
-                      Debug.LogError(e);
+                      Debug.LogError("MobiledgeX: " + e);
```
franklin-huang-mobiledgex, line 357: you can make the "MobiledgeX: " a static string or something, so you don't have to hardcode everytime

ah1053: Will do next update, Thanks for point this out

MobiledgeX/Runtime/Scripts/MobiledgeXIntegration.cs
```diff
@@ -203,124 +172,123 @@ public async Task<bool> VerifyLocation()
...
-                Debug.Log("Exceptions, or app not found. Not Registered!");
-                return null;
+                Debug.LogError("MobiledgeX: Make sure your credentials in MobiledgeX Settings are the same as on MobiledgeX Console.\n check the exceptions for more info.");
+                throw new RegisterClientException("RegisterClient failed");
             }
             else
             {
                 FindCloudletReply reply;
                 Debug.Log("Finding Cloudlet...");
                 reply = await FindCloudlet();
```
franklin-huang-mobiledgex, line 252: I don't think you should reimplement this. We already have FindCloudlet with Performance mode in C# SDK

MobiledgeX/Runtime/Scripts/MobiledgeXIntegration.cs
```diff
@@ -337,69 +305,109 @@ public async Task<String> GetURI()
...
-            Debug.Log("Unable to find ip address for local cellular interface.");
-            return false;
-          }
-    
-          return true;
-        }
-        /// <summary>
-        /// Enum for holding GetConnection Protocols (TCP, UDP, HTTP, Websocket)
-        /// </summary>
-        public enum GetConnectionProtocols
```
franklin-huang-mobiledgex, line 388: I think we still need this because TCP and UDP we bind to cellular interface, so the user could have a wifi associated ip and still be edge enabled. Don't remember what we decided though.

ah1053: I am having a problem responding on slack will send you on Google Hangouts

PR #7, opened 2020-05-13
Restructuring Unity Package according to Unity Package Manager

ah1053:
Package.json along with other folders need to be in the root of the project to enable importing in Unity package manager


franklin-huang-mobiledgex:
As long as this structure is the norm for Unity Packages, looks fine to me

PR #9, opened 2020-05-14
EDGECLOUD-2758  & EDGECLOUD-2720 , GetPort and Default App Name

ah1053:
1. AppName defaults to (Unity Product Name) @lgarner 
2. After CheckCredentials passes, FindCloudlet request occurs retrieving mapped ports and listing them as a drop down list in MobiledgeX Settings. @lgarner 
3. MobiledgeXSettings have fields for TCPPorts and UDPPorts (DropDownList)
Screen Shot 2020-05-14 at 10 33 49 AM
4. Removing NetTest variables 
5. EDGECLOUD-2793 (Auth Public Key added to MobiledgeXSettings)  @franklin-huang-mobiledgex


Runtime/Scripts/MobiledgeXIntegration.cs
```diff
@@ -194,8 +189,7 @@ public async Task<ClientWebSocket> GetWebsocketConnection(string path = "", int
             {
                 port = tcpPort;
             }
-             
-            FindCloudletReply findCloudletReply = await me.RegisterAndFindCloudlet(orgName, appName, appVers, location, carrierName);
+            FindCloudletReply findCloudletReply = await me.RegisterAndFindCloudlet(orgName, appName, appVers, location, carrierName, developerAuthToken.Length > 1 ? developerAuthToken : null);
```
franklin-huang-mobiledgex, line 192: Why greater than 1? Shouldn't this default to empty string? Same for line 354

Runtime/Scripts/MobiledgeXSettings.cs
```diff
@@ -1,14 +1,167 @@
...
+        public string fqdn_prefix;
+#if UNITY_EDITOR
+        [ReadOnly]
+#endif
+        public string path_prefix;
+#if UNITY_EDITOR
+        [ReadOnly]
+#endif
+        public int internal_port;
+#if UNITY_EDITOR
```
franklin-huang-mobiledgex, line 64: What are all these UNITY_EDITOR compiler checks for?

Runtime/Scripts/MobiledgeXSettings.cs
```diff
@@ -1,14 +1,167 @@
...
+        {
+            GUI.enabled = false;
+            EditorGUI.PropertyField(position, property, label, true);
+            GUI.enabled = true;
+        }
+    }
+#endif
+    // MatchingEngine AppPort is not Serializable
+    // Protocol enum is used to override LProto easier to read
+    [System.Serializable]
```
franklin-huang-mobiledgex, line 40: Do we know why AppPort is not serializable? Are we missing a field, or was this the C# json problems that Garner was running into?

Runtime/Scripts/MobiledgeXSettings.cs
```diff
@@ -1,14 +1,167 @@
...
+        UDP,
+        Unknown
+    }
+
+     public class MobiledgeXSettings: ScriptableObject
+    {
+        public string orgName;
+        public string appName;
+        public string appVers;
+        public string authPublicKey;
```
franklin-huang-mobiledgex, line 122: is there any way to add an "optional" note next to this? Also, developer may also want uniqueID and uniqueIDType

Runtime/Scripts/MobiledgeXIntegration.cs
ah1053: Screen Shot 2020-05-15 at 2 38 03 PM

ah1053: Fixed Greater than 0, it defaults to null as per the C# , Tested it and its working

Runtime/Scripts/MobiledgeXSettings.cs
ah1053: ReadOnly Attribures for making mapped ports immutable in UnityEditor, the Same Object (MobiledgeXSettings is used during Runtime on builds) so it cant have UnityEditor functions nor attributes.

Runtime/Scripts/MobiledgeXSettings.cs
ah1053: AppPort have the System.Serializable Attribute on the C# SDK, also LProto is less friendly than Protocol, I can use AppPort later if its updated on the SDK

Runtime/Scripts/MobiledgeXSettings.cs
ah1053: Regarding Unique ID and UniqueIDType posted a question for that on the Slack channel,
Regarding making (AuthPublicKey) optional its not included in the Editor Window (used for Setup) its exposed MobiledgeX Settings(Similar to extra config options on Photon Settings)

lgarner: Those are for MEL only, so...no developer should want uniqueID and uniqueIDType, including us outside the SDK runtime.

franklin-huang-mobiledgex: Ok uniqueID discussion continuing in slack

Runtime/Scripts/MobiledgeXSettings.cs
lgarner: You mean the enum being broken?

Runtime/Scripts/MobiledgeXSettings.cs
franklin-huang-mobiledgex: authpublic key should also be in the Set-Up. If a user supplies an authkey when creating their app (in console or mcctl), they must supply a JWT token with orgname, appname, appvers information in that token otherwise RegisterClient will fail and user will not be able to setup

Runtime/Scripts/MobiledgeXSettings.cs
franklin-huang-mobiledgex: The enum broken?

ah1053: No its not ,
1. Lproto is less user friendly.
2. LProto.HTTP & LProto.UNKNOWN , while Ports on console are either TCP or UDP
3. AppPort is in the dll so I can add the Attribute System.Serializable to it (Used to to expose a variable in Unity Inspector and to Include Inspector changes with Unity Hot Loading)

Editor/MobiledgeXEditorWindow.cs
```diff
@@ -194,12 +196,13 @@ private async void SetupWindow()
                   MobiledgeXIntegration.orgName = settings.orgName;
                   MobiledgeXIntegration.appName = settings.appName;
                   MobiledgeXIntegration.appVers = settings.appVers;
-                  MobiledgeXIntegration.tcpPort = settings.TCP_Port;
-                  MobiledgeXIntegration.udpPort = settings.UDP_Port;
-                  progressText = "";
+                  MobiledgeXIntegration.tcpPort = (int)settings.TCP_Port;
+                  MobiledgeXIntegration.udpPort = (int)settings.UDP_Port;
+                progressText = "";
```
lgarner, line 201: nit, space.

PR #10, opened 2020-05-16
EDGECLOUD-2803 

ah1053:
1. Adding MEL (I didnt do anything) just copied the code from MELClient.
2. EditorWindow Change (Remove AssetsDatabase.Refresh() from OnGUI() since it runs every frame the Window is open. 
3. New Dll 2.0.4 (From Garner Branch ).
4. Removed IsEdgeEnabled for now.
5. Removed VerifyLocation from MobiledgeXIntegration, its already in the SDK.
6. Added Location Permission to PlatformIntegrationUtil.cs
7. In MobiledgeXIntegration.cs, ConfigureMobiledgeXSettings always takes location argument.
8. Seperating ConfigureMobiledgeXSettings(Setting OrgName,AppName,AppVers) from Location and CarrierName.
9. Mel.aar added to the project (with its meta file)
10. EditorScript have functionality to move mel.aar in the Assets folder under(Plugins/MobiledgeX/Android/mel.aar)


Runtime/Scripts/MobiledgeXIntegration.cs
```diff
@@ -111,13 +111,12 @@ public async Task<Loc> GetLocationFromDevice()
         public async Task<bool> Register()
         {
             await ConfigureMobiledgeXSettings(LocationNeeded: false);
```
franklin-huang-mobiledgex, line 113: LocationNeeded should be true

franklin-huang-mobiledgex: Not sure why editor works...

Runtime/Scripts/MobiledgeXIntegration.cs
```diff
@@ -302,50 +253,11 @@ public async Task<String> GetURI(LProto protocol = LProto.L_PROTO_TCP, int port
                 }
             }
             return uri;
-        }
-
-        /// <summary>
-        /// Wether Edge is Enabled on the device or not, Edge requires connections to run over cellular interface
-        /// </summary>
-        /// <returns> boolean value </returns>
-        public bool IsEdgeEnabled()
```
franklin-huang-mobiledgex, line 311: Removing IsEdgeEnabled entirely?

Editor/MobiledgeXEditorWindow.cs
```diff
@@ -98,7 +98,7 @@ private void Awake()
           }
           private void OnGUI()
           {
-              AssetDatabase.Refresh();
+
```
lgarner, line 101: ok, this as you mentioned, stops blasting Cpu to save every GUI update call.

Runtime/Scripts/LocationService.cs
```diff
@@ -60,8 +51,11 @@ public static async Task<LocationInfo> UpdateLocation()
             {
                 // For UNITY_ANDROID, you will need to create an app specific UI to request access to sensitive permissions.
 #if UNITY_ANDROID
+        if(!Permission.HasUserAuthorizedPermission(Permission.CoarseLocation) || !Permission.HasUserAuthorizedPermission(Permission.FineLocation))
+                {
+                    Permission.RequestUserPermission(Permission.FineLocation);  
+                }
```
lgarner, line 57: 👍

Runtime/Scripts/MelMessaging.cs
```diff
@@ -0,0 +1,112 @@
+﻿/**
```
lgarner, line 1: Great! We need to manually update mel.aar binary as needed in the GitHub so developers pull from live repo with the binary.

Runtime/Scripts/MobiledgeXIntegration.cs
```diff
@@ -110,14 +109,13 @@ public async Task<Loc> GetLocationFromDevice()
         /// <returns>Boolean Value</returns>
         public async Task<bool> Register()
         {
-            await ConfigureMobiledgeXSettings(LocationNeeded: false);
-
+            await ConfigureMobiledgeXSettings();
+            me.SetMelMessaging(new MelMessaging(appName));
```
lgarner, line 113: Hi, As we just discussed. Init once, the appName shouldn't change during runtime. MelMessaging happens to be mostly static, but that depends on the class loader if that's true. Should move out to some Init outside.

Runtime/Scripts/MobiledgeXIntegration.cs
```diff
@@ -129,53 +127,6 @@ public async Task<FindCloudletReply> FindCloudlet()
             return reply;
         }
 
-        /// <summary>
```
lgarner, line 132: I sort of still want it, perhaps as an example to be customized to mean something to a particular developer. Likely do so once we implement events (Location too far, etc).

Runtime/Scripts/MobiledgeXIntegration.cs
lgarner: I think we still need it, even if not called (it's used, should be used) only just after FindCloudlet().

Runtime/Scripts/MobiledgeXIntegration.cs
```diff
@@ -357,20 +269,8 @@ public async Task ConfigureMobiledgeXSettings(bool LocationNeeded = true)
...
-            // Location is ephemeral, so retrieve a new location from the platform. May return 0,0 which is
-            // technically valid, though less likely real, as of writing.
-
-            if (LocationNeeded)
-            {
-                location = await GetLocationFromDevice();
-            }
-
+            // Get Location From Device
+            location = await GetLocationFromDevice();
```
lgarner, line 273: Location should be just before FindCloudlet, if actually empty, FindCloudlet shouldn't even be called as it's a required parameter. Complain loudly, and often if it's a developer task.

Runtime/Scripts/MobiledgeXIntegration.cs
```diff
@@ -77,9 +77,23 @@ public void useWifiOnly(bool useWifi)
...
         /// </summary>
         /// <returns></returns>
-        public string GetCarrierName()
+        public void GetCarrierName()
         {
-            return me.carrierInfo.GetMccMnc();
+            // Getting Carrier Name
+            if (me.useOnlyWifi)
+            {
+                carrierName = MatchingEngine.wifiCarrier;
```
franklin-huang-mobiledgex, line 85: we want to return "" instead of "wifi"

Runtime/Scripts/MobiledgeXIntegration.cs
```diff
@@ -110,72 +122,26 @@ public async Task<Loc> GetLocationFromDevice()
         /// <returns>Boolean Value</returns>
         public async Task<bool> Register()
         {
-            await ConfigureMobiledgeXSettings(LocationNeeded: false);
-
+            location = await GetLocationFromDevice();
+            GetCarrierName();
```
franklin-huang-mobiledgex, line 126: Register does not need CarrierName

Runtime/Scripts/MobiledgeXIntegration.cs
```diff
@@ -77,9 +77,23 @@ public void useWifiOnly(bool useWifi)
         /// Returns the MccMnc (Mobile Country Code Mobile Network Code)
         /// </summary>
         /// <returns></returns>
-        public string GetCarrierName()
+        public void GetCarrierName()
```
franklin-huang-mobiledgex, line 80: Why void instead of string? We can get rid of the carrierName class variable if you return string instead

franklin-huang-mobiledgex: Or I kind of liked it back in the ConfigureMobiledgeXSettings function. Why the move?

ah1053: @franklin-huang-mobiledgex, I spoke with @lgarner and he mentioned that it would be better if setting MobiledgeXIntegration variables happen independent of Location and CarrierName would be better

lgarner: GetCarrierName ought to still return carrierName, otherwise it's updateCarrierName().

Runtime/Scripts/MobiledgeXIntegration.cs
ah1053: Fixed, Location is separated from ConfigureMobiledgeXSettings now

Editor/MobiledgeXEditorWindow.cs
ah1053: its added (After AddMobiledgeXPlugins )

Runtime/Scripts/MelMessaging.cs
ah1053: No need to manually added it, I added mel.aar(1.0.4) and their is functionality in the EditorScript to move it from the package to Plugins/MobiledgeX/Android/mel.aar

Runtime/Scripts/MobiledgeXIntegration.cs
ah1053: Fixed

Runtime/Scripts/MobiledgeXIntegration.cs
ah1053: Will be added later

Runtime/Scripts/MobiledgeXIntegration.cs
ah1053: Fixed

Runtime/Scripts/MobiledgeXIntegration.cs
ah1053: Added back

Runtime/Scripts/MobiledgeXIntegration.cs
ah1053: Done, MelMessaging Initialization is in the MobiledgeXIntegration Constructor

Runtime/Scripts/MobiledgeXIntegration.cs
ah1053: Done

Runtime/Scripts/MobiledgeXIntegration.cs
ah1053: NameUpdated

Runtime/Scripts/MobiledgeXIntegration.cs
```diff
@@ -110,72 +123,24 @@ public async Task<Loc> GetLocationFromDevice()
         /// <returns>Boolean Value</returns>
         public async Task<bool> Register()
         {
-            await ConfigureMobiledgeXSettings(LocationNeeded: false);
-
+            location = await GetLocationFromDevice();
```
franklin-huang-mobiledgex, line 126: Don't need location in Register either

lgarner:
I think I recognized much of this code as already merged and obsolete (this PR is not between branch and master, but ports to GarnersMel, the later of which is now master).

franklin-huang-mobiledgex:
@lgarner Should we close this PR?

PR #15, opened 2020-05-22
remove cell id crashing code

franklin-huang-mobiledgex:
Did a quick look over JNI classes that we use and it looks like only CellInfo information is blacklisted. 

Leaving this commented out instead of removing it entirely, unless there are objections.


lgarner:
Might be worth adding a tiny comment that Some JNI code is blacklisted in newer Android OS builds. Not enough to suggest someone actually to enable such code (!), but in case we need it for older phones,

PR #16, opened 2020-05-26
add .meta files that were deleted

franklin-huang-mobiledgex:
minor pr. Git was giving me grief about these files each time I pulled

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,7 @@ ExportedObj/
 *.opendb
 *.VC.db
 *.DS_Store
+
 # Unity3D generated meta files
 *.pidb.meta
 *.pdb.meta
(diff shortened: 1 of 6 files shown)


franklin-huang-mobiledgex:
The meta should only change when I replace the dll. However, I accidently deleted the other files for some reason, so I think these should be new.

franklin-huang-mobiledgex:
Although, it seems like there are no issues in Master (or in this branch) when I import via git url, so not sure what is going on

lgarner:
Let me know, I've bumped meta files for the current binary usage.

lgarner:
I've been not updating the SDK at all for binary file changes, and it seems fine. The UIDs shouldn't be changing anyway...it's how Unity knows that when files moved aroun, they are the same files. Regenerate them, and it no longer knows how to do that.

PR #19, opened 2020-05-28
EDGECLOUD-2903: Unity SDK - Integration script: Call to VerifyLocation() is failing

franklin-huang-mobiledgex:
1) Took Garner's MobiledgeXLocationService from Companion app and copied into LocationService.cs. isEnabledByUser is now checked on the main thread and LocationInfo is updated on each Update() (when LocationServiceStatus is running). @LeonAdams This should fix the issues you were seeing in your tests. 

(@lgarner Just so that I'm clear, the reason isEnabledByUser has to be checked on the main thread is: It is a property, which can hide function calls and all UnityEngine functions must be called from the main thread (for some reason...))

2) Added Unit tests for FindCloudlet and VerifyLocation 

3) Changing settings.orgName does not have any affect on the subsequent API calls. We have to change MobiledgeXIntegration.orgName in order to change the parameters for Register(), FindCloudlet(), etc.

### DIFF
--- a/Tests/Runtime/MobiledgeX_RuntimeTests.cs
+++ b/Tests/Runtime/MobiledgeX_RuntimeTests.cs
@@ -34,7 +34,7 @@ namespace MobiledgeX
             }
             else
             {
-                settings = Resources.Load<MobiledgeXSettings>("MobiledgeXSettings");
+                // settings = Resources.Load<MobiledgeXSettings>("MobiledgeXSettings");
                 integration = new MobiledgeXIntegration();
                 testLocation = new Loc
                 {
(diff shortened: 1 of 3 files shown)


Tests/Runtime/MobiledgeX_RuntimeTests.cs
```diff
@@ -61,24 +61,57 @@ public void Clean()
...
             });
 
+            Debug.Log("result of registerClient is " + task.Result);
             Assert.True(task.Result);
         }
 
+        [Test]
+        [TestCase("MobiledgeX", "MobiledgeX SDK Demo", "2.0")]
+        public void FindCloudlet(string orgName, string appName, string appVers)
+		{
```
brucearmstrong, line 79: Nit: Indentation here and for VerifyLocation.

lgarner:
> (@lgarner Just so that I'm clear, the reason isEnabledByUser has to be checked on the main thread is: It is a property, which can hide function calls and all UnityEngine functions must be called from the main thread (for some reason...))

Yes, Unity is primarily a giant event system of MonoBehavior GameObjects, and background Async threading needs a nice clear hand off point of completed data to synchronous code that doesn't violate the rules of MonoBehavior (or freeze the UI).

PR #20, opened 2020-05-29
Update target binaries

lgarner:
Update to Mel.aar 1.0.7. Update MatchingEngine dll to the latest build 2.0.7. Unit tested in test Unity app.


franklin-huang-mobiledgex:
Does the dll need a new meta file?

lgarner:
> Does the dll need a new meta file?

I'm staring right at it not changed, and new one didn't show up despite using the new DLL. Don't know how relevant that is when the first thing Setup does is move/delete files...

franklin-huang-mobiledgex:
I think when you delete/replace a file, you need to delete the meta file and then when you add the new file Unity will generate a new meta file (with a new guid)

PR #21, opened 2020-06-04
2.0.7 update

franklin-huang-mobiledgex:
Quite a large PR:

1. Bump dll to 2.0.7
2. In MobiledgeXIntegration.cs, DME wrappers (ie. RegisterClient, FindCloudlet, VerifyLocation) all return a Task<bool>. Useful results from those API calls are stored in MobiledgeXIntegration class so that developers don't have to look at FindCloudletReply or other objects. 
3. Better exception handling for Wrappers, so that developers only have to catch relevant exceptions
4. Add GetAppPort function. This allows the developer to choose which AppPort they want or if the developer does not supply anything it grabs the first AppPort in the dictionary
5. Modfiy GetUrl and add GetPort and GetHost functions for developers to use
6. Add a MobiledgeXIntegrationConfig file. This has functions for developer to configure MobiledgeXIntegration (settings, UseWifiOnly, FindCloudletMode)
7. Add a MobiledgeXIntegrationHelper file. These are internal helper functions for MobiledgeXIntegration.
8. Lastly, I am encouraging a new workflow for developers using Integration. I didn't like the black box of only calling GetWebsocketConnection and this made it difficult to make GetUrl, GetHost and GetPort consistent (developer needs some control). The new workflow is: RegisterAndFindCloudlet -> GetAppPort -> GetConnection/GetUrl/GetHost + GetPort. See MobiledgeXGetWebsocketConnection() in MobiledgeXSocketClient.cs for an example with exception handling


Runtime/Scripts/Example.cs
```diff
@@ -59,12 +61,69 @@ async Task RestExample()
...
+        integration.UseWifiOnly(true);
 #endif
+        // RegisterAndFindCloudlet
+        bool registeredAndFoundCloudlet;
+        try
+        {
+            registeredAndFoundCloudlet = await integration.RegisterAndFindCloudlet();  
+        }
+        catch (RegisterClientException rce)
+		{
```
brucearmstrong, line 73: Indentation looks odd here.

franklin-huang-mobiledgex: hmm weird. will fix these indents

PR #23, opened 2020-06-06
Update to work with MEL. MatchingEngine DLL and mel.aar.

lgarner:
Working MEL with pull at mel-sample-client (mel_unity_sdk).

- Working within limits, default mccmnc + appInst selected cannot work with GCP, as it is ambiguous between appInsts. Falls back to Proximity/Performance FindCloudlet.

- MEC cannot resolve the incomplete edge url, and will not return DNS resolve to it (probably a good thing).


lgarner:
This branch works directly from git in the mel-sample-app. Note that there's no meta file updates.

PR #24, opened 2020-06-09
EDGECLOUD-2969: Unity SDK - MobiledgeX Setup throwing a helper function exception instead of app not found

franklin-huang-mobiledgex:
RegisterClient absorbs HttpException and CheckCredentials() did not catch RegisterClientExcpetion

### DIFF
--- a/Editor/MobiledgeXEditorWindow.cs
+++ b/Editor/MobiledgeXEditorWindow.cs
@@ -330,14 +330,6 @@ namespace MobiledgeX
                     Debug.LogError("No Mapped Ports for your application backend");
                 }
                 return checkResult;
-
-
-            }
-            catch (HttpException httpe) // HTTP status, and REST API call error codes.
-            {
-                // server error code, and human readable message:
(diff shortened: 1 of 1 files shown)


Editor/MobiledgeXEditorWindow.cs
```diff
@@ -330,24 +330,21 @@ public async Task<bool> CheckCredentials()
...
-                return false;
             }
             catch (HttpRequestException httpre)
             {
                 clog("MobiledgeX: RegisterClient HttpRequest Exception", httpre.Message + "\nStack Trace: " + httpre.StackTrace, true);
                 return false;
             }
-            catch (FindCloudletException findCloudletException)
+            catch (RegisterClientException rce) 
+			{
```
lgarner, line 340: Nit: whitepace weird.

franklin-huang-mobiledgex: RegisterClientException. RegisterClient() catches HttpException and then throws a RegisterClientException with status, message, error code

PR #27, opened 2020-06-11
EDGECLOUD-2980: Port Mapping Error Preventing Unity Builds

franklin-huang-mobiledgex:
Comment out AppPort UI code. To be fixed later


franklin-huang-mobiledgex:
> OK. To be sure, It's not connected to any _app_ code in the rest of MobiledgeXIntegration, right? (I guess it compiles, so probably so).

Nope. Everything compiles and I ran the ConnectionExample and things work there

PR #29, opened 2020-06-11
Bump to v2.1.0

lgarner:
Just a version string, not a release tag. Merge when ready.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "com.mobiledgex.sdk",
-  "version": "2.0.10",
+  "version": "2.1.0",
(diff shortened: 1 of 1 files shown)


franklin-huang-mobiledgex:
Wait I forget what we agreed on. Is this version the C# SDK version or the Unity SDK version?

lgarner:
> Wait I forget what we agreed on. Is this version the C# SDK version or the Unity SDK version?

Both actually, but this is the string we show to the Unity UI.

franklin-huang-mobiledgex:
ok. so c# SDK also needs the 2.1 version bump

PR #33, opened 2020-06-16
EDGECLOUD-3019: Check for Unity Editor before calling location services functions in Unity

franklin-huang-mobiledgex:
1. UnityEditor compiler check before calling LocationServices. If in editor, no location will be returned, so we return some random location immediately. This fixes TestRunner bugs for FindCloudlet and VerifyLocation

2. Remove Connect(string path) function which had redundant GetConnection workflow
- This function was confusing and didn't allow developers to choose ports
3. Adjust Example.cs accordingly


franklin-huang-mobiledgex:
I don't think it matters too much. Its whatever the developer wants to name it. I like mxi because its the shortest

Runtime/Scripts/MobiledgeXIntegrationHelper.cs
```diff
@@ -131,15 +131,23 @@ private Loc GetLocationFromDevice()
         {
             // Location is ephemeral, so retrieve a new location from the platform. May return 0,0 which is
             // technically valid, though less likely real, as of writing.
+#if UNITY_EDITOR
+            Debug.Log("MobiledgeX: Cannot Get location in Unity Editor. Returning dummy location.");
+            return new Loc{
+                latitude = 37.3382082,
+                longitude = -121.8863286
```
lgarner, line 138: Oh, just getting visions of the whole (long, lat), versus (lat, long) debate. But here, they're named them at least.

Runtime/Scripts/MobiledgeXIntegrationHelper.cs
```diff
@@ -131,15 +131,23 @@ private Loc GetLocationFromDevice()
...
+            };          
+#else
             Loc loc = LocationService.RetrieveLocation();
-            // If in UnityEditor, 0f and 0f are hard zeros as there is no location service.
+            // 0f and 0f are hard zeros if no location service.
```
lgarner, line 142: I would actually like to return 0's here, if that's going to be part of the Integration layer.

franklin-huang-mobiledgex: I get this when plugging in 0,0: FindCloudletException : FindCloudlet Exception: Missing GpsLocation, HTTP StatusCode: BadRequest, API ErrorCode: 3

franklin-huang-mobiledgex: May be a dme error that doesn't allow 0,0

lgarner: Right, how about a configurable fallback up top somewhere instead. It's 99.99% sure to be an error with 0,0, but at least they won't have to use our very specific location.

franklin-huang-mobiledgex: so we fallback to this location struct if using UnityEditor or if LocationServices returns 0,0?

lgarner: Yeah, hope is not use it much outside the editor since there exists an example LocationService with permissions and everything.

franklin-huang-mobiledgex: I think for now: I will let (0,0) slide, but print an error log. Then developer will probably receive the same FindCloudletException as I did and they will see that they are getting (0,0) location

Runtime/Scripts/MobiledgeXIntegrationHelper.cs
```diff
@@ -131,15 +143,22 @@ private Loc GetLocationFromDevice()
...
+            return loc;
+#else
+            loc = LocationService.RetrieveLocation();
+            // 0f and 0f are hard zeros if no location service.
             if (loc.longitude == 0f && loc.latitude == 0f)
             {
-                // Likely not in the ocean. We'll chose something for demo FindCloudlet purposes:
-                loc.longitude = -121.8863286;
-                loc.latitude = 37.3382082;
+                Debug.ErrorLog("LocationServices returned a location at (0,0)");
```
lgarner, line 158: 👍

PR #42, opened 2020-06-24
Bump csharp binary

lgarner:
Support MEL -> none or no UID, but still a particular phone manufacturer by using newly suggested ANDROID_ID. Has some privacy implications. The developer should be informed if on such a phone configuration to notify user.

2.1.1 should land first.


lgarner:
Moving to staging_2.1.x

PR #43, opened 2020-06-24
Staging 2.1.x

lgarner:
Support MEL -> none or no UID, but still a particular phone manufacturer by using newly suggested ANDROID_ID. Has some privacy implications. The developer should be informed if on such a phone configuration to notify user.

2.1.1 nuget should land first in Nuget-Development (This actually includes the fix for carrierName = null in FindCloudlet)


Runtime/Scripts/NetInterfaceIntegration.cs
```diff
@@ -101,9 +101,12 @@ public bool HasCellular()
       NetworkInterface[] netInterfaces = GetInterfaces();
```
brucearmstrong, line 101: Copyright date.

lgarner: ?

PR #44, opened 2020-06-26
Fix typo, add edge default network path check independent of UseWiFiO…

lgarner:
2.1.1 in nuget-development (current version of it). It has an expanded default list of interfaces due to dynamic numbering of interface names for cellular.

The HasWiFi() and HasCellular() checks checks for both OperationalStatus.UP and looks for an IP address in a heuristic manner. It is best effort only. Unfortunately, it is apparently broken in Unity SDK for Cellular interfaces at least. Status is always Unknown. WiFi status is correct, but it is also IPv4 only on my test home network.

NetTest may be used in the future to come up with a better answer, but OperationStatus should be doing the same if it works correctly.

The C# SDK DLL version remains the same. Unity SDK version overrides that version with its own default.

All this supports a single call on that answers this question independent of the supported protocols in the Unity SDK:
"Do we have an edge Netowrk Path available? " + mxi.IsNetworkDataPathEdgeEnabled());

By inspection, this should be correct for IOS as well (ipv4 triggers true), even if the IOS bug says OperationStatus == Unknown.


Runtime/Scripts/MobiledgeXIntegrationHelper.cs
```diff
@@ -207,7 +232,7 @@ private bool IsEdgeEnabled(GetConnectionProtocol proto)
             }
```
ah1053, line 233: Are you planning to add IsNetworkDataPathEdgeEnabled to IsEdgeEnabled?
 if (!IsNetworkDataPathEdgeEnabled())
            {
                Debug.Log(".....");
                return false;
            }

lgarner: I'm not changing it yet, as it's primarily internal, and asking a slightly different question. It does benefit from being able to get the correct status however, subject to limits of the heuristic used inside HasWiFi/Cellular.

This one at least, retrieves a simple status, that's composable to answer different questions.

Runtime/Scripts/NetInterfaceIntegration.cs
```diff
@@ -99,29 +99,77 @@ public string GetIPAddress(string sourceNetInterfaceName, AddressFamily addressf
...
+          }
+          else if (GetIPAddress(cellularName, AddressFamily.InterNetworkV6) != null)
+          {
+            // No-op. Every interface has IpV6.
+          }
+          else if (GetIPAddress(cellularName, AddressFamily.InterNetwork) != null)
+          {
+            foundByIp = true;
+          }
+          if (foundByIp)
```
franklin-huang-mobiledgex, line 131: can't we just return foundByIp?

lgarner: I suspect we might be expanding this if we find more edge cases.

PR #45, opened 2020-06-26
(Typo) proto

ah1053:


Runtime/Scripts/Example.cs
```diff
@@ -135,7 +135,7 @@ private async Task<String> MobiledgeXIntegrationWorkflow(MobiledgeXIntegration i
         string url;
         try
         {
-            url = integration.GetUrl("http");
+            url = integration.GetUrl(proto);
```
lgarner, line 138: Okay, upper layer needs to pass in what to prepend.

PR #46, opened 2020-06-26
link.xml Update to support .NET 4.x

ah1053:
Garner Solution to support .NET 4.X ([ApiCompatibilityLevel](https://docs.unity3d.com/ScriptReference/ApiCompatibilityLevel.html))
This solution will preserve the two assemblies from [stripping](https://docs.unity3d.com/2018.3/Documentation/Manual/ManagedCodeStripping.html) during the build
([link](https://docs.microsoft.com/en-us/visualstudio/cross-platform/unity-scripting-upgrade?view=vs-2019))


lgarner:
Actually, was planning the merge brach to be in staging_2.1.x to keep them separate.

PR #48, opened 2020-06-29
Add edge network check

lgarner:
Additional fix in MatchingEngine DLL for GetAvailableInterfaces for GetLocalIP.


Runtime/Scripts/MobiledgeXIntegrationHelper.cs
```diff
@@ -169,6 +170,30 @@ private void UpdateCarrierName()
...
 
+        /// <summary>
+        /// Checks whether the default netowrk data path Edge is Enabled on the device or not, Edge requires connections to run over cellular interface.
+        /// This status is independent of the UseWiFiOnly setting.
+        /// </summary>
+        /// <returns>bool</returns>
+        public bool IsNetworkDataPathEdgeEnabled() {
+            string wifiIpV4 = null;
+            string wifiIpV6 = null;
+            string cellIp = null;
```
ah1053, line 181: Are you using cellIp?

lgarner: no, good catch, not at this level.

PR #49, opened 2020-06-30
IsRoaming function for IOS

franklin-huang-mobiledgex:
Similar logic to https://github.com/mobiledgex/MatchingEngineSDK/pull/74

1. Added an IsRoaming function to CarrierInfoClass for now (will add to CarrierInfoInterface in C# SDK and add this check to GenerateDMEHost and GetCarrierName if this is approved)
2. Added _getISOCountryCodeFromGPS(), _getISOCountryCodeFromCarrier(), and _convertGPSToISOCountryCode(double longitude, double latitude) functions to PlatformIntegration.m file
- _convertGPSToISOCountryCode(double longitude, double latitude) gets the ISO Country code from gps location asynchronously. Once done, it stores the ISO Country code in an NSString* class variable (isoCountryCode). 
- _getISOCountryCodeFromGPS() will grab that class variable
- _getISOCountryCodeFromCarrier() gets the ISO cc from the carrier (CoreTelephony)
3. Added GetISOCountryCodeFromGPS(), GetISOCountryCodeFromCarrier(), ConvertGPSToISOCountryCode(double longitude, double latitude) to CarrierInfoClass. These correspond to each of the functions in PlatformIntegration.m
- These functions are used by IsRoaming to compare ISO Country Code from gps location with ISO Country Code from the carrier
- ConvertGPSToISOCountryCode(double longitude, double latitude) returns a Task<string>. This task polls the isoCountryCode variable by calling GetISOCountryCodeFromGPS() until it is not null or an empty string.
- GetISOCountryCodeFromGPS() grabs the isoCountryCode variable from PlatformIntegration.m
- GetISOCountryCodeFromCarrier() gets the ISO cc from the carrier
4. Added some sample code to show how to use this IsRoaming function. The first try/catch tests a San Jose GPS location and the second try/catch tests some location in the Atlantic that will not have an ISO country code.

### DIFF
--- a/Runtime/Plugins/iOS/PlatformIntegration.m
+++ b/Runtime/Plugins/iOS/PlatformIntegration.m
@@ -258,7 +258,7 @@ char* _getISOCountryCodeFromCarrier()
 
     if (@available(iOS 12.1, *))
     {
-        carrier = [networkState.lastCarrier];
+        carrier = networkState.lastCarrier;
     }
     else
     {
@@ -266,9 +266,7 @@ char* _getISOCountryCodeFromCarrier()
         carrier = [netinfo subscriberCellularProvider]; // s for dual SIM?
         NSLog(@"Carrier Name: %@", [carrier carrierName]);
         // Ref counted.
-
-        nsstr = [carrier carrierName];
     }
     NSLog(@"ISO Country code: %@", carrier.isoCountryCode);
-    return convertToCStr(carrier.isoCountryCode);
+    return convertToCStr([carrier.isoCountryCode UTF8String]);
 }
--- a/Runtime/Plugins/iOS/PlatformIntegration.m
+++ b/Runtime/Plugins/iOS/PlatformIntegration.m
@@ -260,7 +260,6 @@ void _convertGPSToISOCountryCode(double longitude, double latitude)
          if(placemarks && placemarks.count > 0)
          {
              CLPlacemark *placemark= [placemarks objectAtIndex:0];
-
              isoCountryCode = [placemark ISOcountryCode];
          }
      }];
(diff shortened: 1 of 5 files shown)


Runtime/Plugins/iOS/PlatformIntegration.m
```diff
@@ -243,5 +245,47 @@ unsigned int _getCellID()
...
+    isoCountryCode = NULL;
+    CLLocation *location = [[CLLocation alloc] initWithLatitude:latitude longitude:longitude];
+    CLGeocoder *geocoder = [[CLGeocoder alloc] init];
+    [geocoder reverseGeocodeLocation:location completionHandler:^(NSArray *placemarks, NSError
+    *error)
+     {
+         if(placemarks && placemarks.count > 0)
+         {
+             CLPlacemark *placemark= [placemarks objectAtIndex:0];
+
```
ah1053, line 263: nit space

Runtime/Scripts/CarrierInfoIntegration.cs
```diff
@@ -405,6 +428,81 @@ public ulong GetCellID()
...
+    private string GetISOCountryCodeFromCarrier()
+    {
+      string isoCC = null;
+      if (Application.platform == RuntimePlatform.IPhonePlayer)
+      {
+        isoCC = _getISOCountryCodeFromCarrier();
+      }
+      return isoCC;
+    }
+
```
lgarner, line 505: OK. IOS only.

Runtime/Scripts/Example.cs
```diff
@@ -64,6 +64,27 @@ private void Update()
     async Task RestExample()
     {
         MobiledgeXIntegration mxi = new MobiledgeXIntegration();
+        CarrierInfoClass ci = new CarrierInfoClass();
+
+        try
```
lgarner, line 69: Might have missed it. Roaming interface is for IOS only for the moment.

lgarner: Also, use staging_2.1.x.

franklin-huang-mobiledgex: ya i need to add compiler directives. thanks

lgarner:
Thanks. There may be a need for "isRoaming" in C# like there Is in the other SDKs later, but IOS needs it first as it only knows the home carrier otherwise.

PR #51, opened 2020-07-01
Update appport for mel

lgarner:
This is a somewhat optional change for ease of use. If using the more advanced multiport/non-dedicated utility functions, allow MEL (dedicated server only) to work in those conditions.

- Some print cleanup.

### DIFF
--- a/Runtime/Scripts/MobiledgeXIntegration.cs
+++ b/Runtime/Scripts/MobiledgeXIntegration.cs
@@ -84,6 +84,8 @@ namespace MobiledgeX
         AppPort[] latestAppPortList = null;
         Location fallbackLocation = new Location(-121.8863286, 37.3382082);
 
+        MelMessaging melMessaging;
+
         /// <summary>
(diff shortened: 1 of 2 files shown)


Runtime/Scripts/MobiledgeXIntegration.cs
```diff
@@ -225,6 +233,46 @@ public AppPort GetAppPort(LProto proto, int port = 0)
...
+        private bool AppPortForMel(FindCloudletReply reply, LProto proto, int defaultPort)
+        {
+            if (IsNetworkDataPathEdgeEnabled() && melMessaging.IsMelEnabled())
+            {
+                if (reply.ports.Length > 1)
+                {
+                    throw new Exception("MobiledgeX: Unexpected Port length for MEL mode.");
+                }
+
+                AppPort appPort = reply.ports[0];
```
franklin-huang-mobiledgex, line 248: Does this appPort need to go back into reply.ports after it is modified?

lgarner: In place modify. The original object is empty to begin with.

PR #52, opened 2020-07-01
Versatile WebSocket EDGECLOUD-3142

ah1053:
Versatile WebSocket:
1. Sending Binary and Text Data (2 threads and two queues).
2. Receiving Binary and Text Data ( 1 thread and two queues).
3. Name change, new name is (MobiledgeXWebSocketClient.cs).

Tested on ComputerVision


franklin-huang-mobiledgex:
@lgarner Is there a situation where a Websocket client may want to send String msg and Binary msg to the same host and port?

brucearmstrong:
> @lgarner Is there a situation where a Websocket client may want to send String msg and Binary msg to the same host and port?

The ComputerVision server can accept either binary or text data, both on the same host, port, and URL. For binary, we operate on it as an image. For text, we simply echo the received text back. This is used for a full round trip "ping" time.

Runtime/Scripts/MobiledgeXWebSocketClient.cs
```diff
@@ -108,9 +89,14 @@ public void Send(string message)
...
             //Debug.Log("Message to queue for send: " + buffer.Length + ", message: " + message);
             var sendBuf = new ArraySegment<byte>(buffer);
-
             sendQueue.Add(sendBuf);
         }
+       public async void Send(byte[] binary)
+        {
+            //Debug.Log("Message to queue for send: " + buffer.Length + ", message: " + message);
+            var sendBuf = new ArraySegment<byte>(binary);
+            await ws.SendAsync(sendBuf, WebSocketMessageType.Binary, true /* is last part of message */, CancellationToken.None);
```
franklin-huang-mobiledgex, line 98: Is there no sendQueue for binary data?

ah1053: I wanted to use the MobiledgeXWebSocketClient in the CV app so small change to send, the sending queue is of type string.

Runtime/Scripts/MobiledgeXWebSocketClient.cs
```diff
@@ -108,9 +89,14 @@ public void Send(string message)
             byte[] buffer = encoder.GetBytes(message);
             //Debug.Log("Message to queue for send: " + buffer.Length + ", message: " + message);
             var sendBuf = new ArraySegment<byte>(buffer);
-
             sendQueue.Add(sendBuf);
         }
+       public async void Send(byte[] binary)
```
lgarner, line 94: Line.

Runtime/Scripts/MobiledgeXWebSocketClient.cs
lgarner: The websocket client has independent queues for send and receive, await would freeze the UI sending that binary buffer.

ah1053: @lgarner  @franklin-huang-mobiledgex  I added another queue for binary messages and an independent thread (sendThreadBinary) that is being initialized at the constructor.

franklin-huang-mobiledgex:
This looks better. Should we add a parameter to the Receive function that indicates whether the developer wants binary data or string data? Developer may must want raw bytes instead of the String conversion

Runtime/Scripts/MobiledgeXWebSocketClient.cs
```diff
@@ -149,31 +172,48 @@ public async Task<string> Receive(UInt64 maxSize = MAXREADSIZE)
...
             {
                 //Debug.Log("Awaiting Receive...");
-                result = await Receive();
-                if (result != null && result.Length > 0)
+                response = await Receive();
+                if (response != null && response.Keys.Count > 0)
                 {
                     //Debug.Log("Received: " + result);
-                    receiveQueue.Enqueue(result);
+                    if (response.Keys.First() == WebSocketMessageType.Text)
```
franklin-huang-mobiledgex, line 199: how does the developer set the receive type?

ah1053: The developer doesn't need to set the receive type, The receive type is dedicated by the server.

ah1053: as discussed
 chunkResult is of type WebSocketReceiveResult not byte[]
WebSocketReceiveResult chunkResult >> WebSocketReceiveResult has MessageType field

PR #54, opened 2020-07-07
Cleaning MobiledgeX Package

ah1053:
Some minor changes:
1. DeveloperAuthToken is used in Check Credentials in Unity Editor.
2. Adding missing copyrights.
3. Better Debugging Messages.
4. Removing Ports GUI.
5. Removing unnecessary namespaces (ex. using System.Threading.Tasks; )


Runtime/Scripts/LocationService.cs
```diff
@@ -110,29 +100,27 @@ public static Loc RetrieveLocation()
     {
       if (Input.location.status != LocationServiceStatus.Running)
       {
-        UnityEngine.Debug.Log("Warning: Location Status must be in running state to get a valid location! Current Status: " + Input.location.status);
+        Debug.LogWarning("MobiledgeX: Location Status must be in running state to get a valid location! Current Status: " + Input.location.status);
       }
       LocationInfo locationInfo = Input.location.lastData;
-      UnityEngine.Debug.Log("Location Info: [" + locationInfo.longitude + "," + locationInfo.latitude + "]");
+      Debug.Log("Location Info: [" + locationInfo.longitude + "," + locationInfo.latitude + "]");
```
lgarner, line 106: Maybe we don't have to print. The location use is just to trigger the Permissions dialog. Might still be useful though.

Runtime/Scripts/LocationService.cs
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright 2019 MobiledgeX, Inc. All rights and licenses reserved.
+ * Copyright 2018-2020 MobiledgeX, Inc. All rights and licenses reserved.
```
franklin-huang-mobiledgex, line 2: good catch

Runtime/Scripts/CarrierInfoIntegration.cs
```diff
@@ -358,6 +356,13 @@ public string GetMccMnc()
 
     public ulong GetCellID()
     {
+      /*
+       * The following code is commented out to prevent Android JNI blacklist crashes.
+       * As of Android API 28, CellInfo interfaces and class reflection through JNI are not allowed.
```
vmohanx, line 361: instead of leaving it commented, can we detect what the android version is with something like this : 
```
   using (var version = new AndroidJavaClass("android.os.Build$VERSION")) {
     return version.GetStatic<int>("SDK_INT");
   }
}
```
from : https://answers.unity.com/questions/976506/how-to-get-andriod-api-level.html

Runtime/Scripts/LocationService.cs
vmohanx: I like having it for debug purposes, but not a hard requirement

Runtime/Scripts/CarrierInfoIntegration.cs
ah1053: Good Point, @franklin-huang-mobiledgex  since you worked in this bug, Do u think we should detect the Android SDK Version, then get Cell ID if the SDK Version is below 28.

 *btw we don't need versions below 24 since this the minimum we support.

franklin-huang-mobiledgex: If anybody starts using CellID with Android SDK version 24-27, we can do that. The code change is simple, as we already grab the SDK version, because it takes us down slightly different code paths. But for now it's all consistent when it is commented out. Either way is not a big deal; Android JNI is easily crashable, so leaving it commented out is less risky

ah1053: Ok

PR #55, opened 2020-07-08
EDGECLOUD-2987: IOS SDK: If in roaming mode, use global DME and FindCloudlet Performance Mode.

franklin-huang-mobiledgex:
1. Add IsRoaming check before each DME API call. If IOS device is roaming, useWifiOnly mode
2. Converted GetLocationFromDevice to UpdateLocationFromDevice (no return value, just updates location global variable to match UpdateCarrierName function)
3. Wrap UpdateLocationFromDevice, IsRoaming, and UpdateCarrierName into one function: UpdateLocationAndCarrierInfo. 
- This function will be called before each DME API to make sure location is updated, the ios roaming check is called to check if we need to useWifiOnly mode, and carrier name is updated
- RegisterClient requires this call also because we need the Roaming check, which requires location

### DIFF
--- a/Runtime/Scripts/MobiledgeXIntegrationHelper.cs
+++ b/Runtime/Scripts/MobiledgeXIntegrationHelper.cs
@@ -69,6 +69,8 @@ namespace MobiledgeX
             Debug.Log("MobiledgeX: AppName: " + req.app_name);
             Debug.Log("MobiledgeX: AppVers: " + req.app_vers);
 
+            await UpdateLocationAndCarrierInfo();
+
             RegisterClientReply reply;
             try
             {
@@ -107,8 +109,7 @@ namespace MobiledgeX
                 throw new FindCloudletException("Last RegisterClient was unsuccessful. Call RegisterClient again before FindCloudlet");
             }
 
-            location = GetLocationFromDevice();
-            UpdateCarrierName();
+            await UpdateLocationAndCarrierInfo();
 
             FindCloudletRequest req = matchingEngine.CreateFindCloudletRequest(location, carrierName);
             FindCloudletReply reply;
@@ -120,7 +121,6 @@ namespace MobiledgeX
             {
                 throw new FindCloudletException("FindCloudlet Exception: " + httpe.Message + ", HTTP StatusCode: " + httpe.HttpStatusCode + ", API ErrorCode: " + httpe.ErrorCode + "\nStack: " + httpe.StackTrace);
             }
-
 
             if (reply == null)
             {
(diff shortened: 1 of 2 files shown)


Runtime/Scripts/MobiledgeXIntegrationHelper.cs
```diff
@@ -69,6 +69,8 @@ public async Task<bool> Register()
             Debug.Log("MobiledgeX: AppName: " + req.app_name);
             Debug.Log("MobiledgeX: AppVers: " + req.app_vers);
 
+            await UpdateLocationAndCarrierInfo();
```
ah1053, line 72: Do we need UpdateLocationAndCarrierInfo() in RegisterClient ?

Runtime/Scripts/MobiledgeXIntegrationHelper.cs
```diff
@@ -139,26 +139,38 @@ public async Task<bool> FindCloudlet()
...
 #if UNITY_EDITOR
             Debug.Log("MobiledgeX: Cannot Get location in Unity Editor. Returning fallback location. Developer can configure fallback location with SetFallbackLocation");
-            loc.longitude = fallbackLocation.Longitude;
-            loc.latitude = fallbackLocation.Latitude;
-            return loc;
+            location.longitude = fallbackLocation.Longitude;
+            location.latitude = fallbackLocation.Latitude;
 #else
-            loc = LocationService.RetrieveLocation();
+            location = LocationService.RetrieveLocation();
```
ah1053, line 168: better naming, Thanks

Runtime/Scripts/MobiledgeXIntegrationHelper.cs
```diff
@@ -107,8 +109,7 @@ public async Task<bool> FindCloudlet()
                 throw new FindCloudletException("Last RegisterClient was unsuccessful. Call RegisterClient again before FindCloudlet");
             }
 
-            location = GetLocationFromDevice();
-            UpdateCarrierName();
+            await UpdateLocationAndCarrierInfo();
```
ah1053, line 112: I am a big fan of consolidation, However a small reminder the [next feature we ll have](https://mobiledgex.atlassian.net/browse/EDGECLOUD-3157) exposing Location Parameter in FindCloudlet.
UpdateLocationAndCarrierInfo(latitude, longtiude) VS UpdateLocation(latitude, longtiude) + UpdateCarrierName()

Runtime/Scripts/MobiledgeXIntegrationHelper.cs
franklin-huang-mobiledgex: Yup. I need the roaming check before RegisterClient in order to set the correct DME (if roaming -> useWifiOnly). And the roaming check requires location

Runtime/Scripts/MobiledgeXIntegrationHelper.cs
franklin-huang-mobiledgex: That feature shouldn't affect the consolidation. I think that feature will just override the UpdateLocation call if the developer provides a long/lat

PR #59, opened 2020-07-13
2.1.2 version string.

lgarner:
2.1.2 version string.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "com.mobiledgex.sdk",
-  "version": "2.1.1",
+  "version": "2.1.2",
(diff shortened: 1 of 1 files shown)


lgarner:
> Can we add the 2.1.2 C# dll to this pr too? Keep the string change and dll update in the same place

Yes, this makes sense. I can update. Need to build C# SDK first of course.

lgarner:
> > Can we add the 2.1.2 C# dll to this pr too? Keep the string change and dll update in the same place
> 
> Yes, this makes sense. I can update. Need to build C# SDK first of course.

Updated binary.

PR #60, opened 2020-07-14
if roaming check fails, assume not roaming

franklin-huang-mobiledgex:
Wasn't catching the exceptions thrown in IsRoaming. If there is a problem, assume that device is not roaming so that we can continue with DME calls.


Runtime/Scripts/MobiledgeXIntegrationHelper.cs
```diff
@@ -203,11 +203,14 @@ private void UpdateCarrierName()
         /// <returns>bool</returns>
         public async Task<bool> IsRoaming()
         {
+#if !UNITY_EDITOR
```
lgarner, line 206: 👍

PR #61, opened 2020-07-14
LocationService

ah1053:
1. **Location Service**
**Android** 
> Check user permission > if permission is not granted yet > ask for permission > wait for user > if denied > throw
> if permission granted > continue with Location Service

**iOS** 

> Proceed with initialize Location > Input.location.start() >( if permission wasn't granted ) iOS permission will pop up>
       > Application will be out of focus > check if Location Request.failed > throw 
                                                               > check if input.lastData ==0 > throw

2.**Removing** GetIPAddress() debugging, it spams the logs

### DIFF
--- a/Runtime/Scripts/LocationService.cs
+++ b/Runtime/Scripts/LocationService.cs
@@ -43,10 +43,10 @@ namespace MobiledgeX
 #else
                 yield break;
 #endif
-            }
+       }
 
-            // Start service before querying location
-            Input.location.Start();
+      // Start service before querying location
+      Input.location.Start();
 
       // Wait until service initializes
       while (Input.location.status == LocationServiceStatus.Initializing && maxWait > 0)
--- a/Runtime/Scripts/LocationService.cs
+++ b/Runtime/Scripts/LocationService.cs
@@ -24,16 +24,22 @@ using System;
 namespace MobiledgeX
 {
 
-  // Unity Location Service, based on the documentation example:
-  public class LocationService : MonoBehaviour
-  {
-
-    public static IEnumerator InitalizeLocationService(int maxWait = 20, bool continuousLocationService = false)
+    // Unity Location Service, based on the documentation example:
+    public class LocationService : MonoBehaviour
     {
-      // First, check if user has location service enabled
-      if (!Input.location.isEnabledByUser)
-      {
-        Debug.Log("MobiledgeX: Location Service disabled by user.");
+        private static bool locationPermissionRejected = false;
+
+        void Awake()
+        {
+            StartCoroutine(LocationServiceFlow());
+        }
+
+        public static IEnumerator InitalizeLocationService(int maxWait = 20, bool continuousLocationService = false)
+        {
+            // First, check if user has location service enabled
+            if (!Input.location.isEnabledByUser)
+            {
+
 #if UNITY_IOS
                 //if isEnabledByUser is false and you start location updates anyway,
                 //the CoreLocation framework prompts the user with a confirmation panel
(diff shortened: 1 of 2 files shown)


Runtime/Scripts/LocationService.cs
```diff
@@ -27,14 +28,23 @@ namespace MobiledgeX
...
         Debug.Log("MobiledgeX: Location Service disabled by user.");
-        yield break;
-      }
+#if UNITY_IOS
+                //if isEnabledByUser is false and you start location updates anyway,
+                //the CoreLocation framework prompts the user with a confirmation panel
+                //asking whether location services should be reenabled.
+                //The user can enable or disable location services altogether from the Settings application
+                //by toggling the switch in Settings>General>LocationServices.
+                //https://docs.unity3d.com/ScriptReference/LocationService-isEnabledByUser.html
```
brucearmstrong, line 43: I appreciate comments like this.

Runtime/Scripts/LocationService.cs
```diff
@@ -72,9 +80,26 @@ public static IEnumerator InitalizeLocationService(int maxWait = 20, bool contin
...
+#else
+          throw new Exception("MobiledgeX: LocationService request Failed, Location Permission is needed for connecting to Edge.");
+#endif
+         }
+    }
+
+    private void Awake()
     {
-      yield return StartCoroutine(InitalizeLocationService());
+         ensurePermissions();
```
franklin-huang-mobiledgex, line 101: nice. i like this here

Runtime/Scripts/LocationService.cs
```diff
@@ -72,9 +80,26 @@ public static IEnumerator InitalizeLocationService(int maxWait = 20, bool contin
       }
     }
 
-    public IEnumerator Start()
+    public static IEnumerator AssureLocation()
+    {
+         if (Input.location.status == LocationServiceStatus.Initializing)
```
franklin-huang-mobiledgex, line 85: I suggest checking if "Input.Location.status != LocationServiceStatus.Running" instead and then checking if LocationServiceStatus is Stopped or Failed, so that we can throw specific exceptions

Runtime/Scripts/LocationService.cs
```diff
@@ -27,14 +28,23 @@ namespace MobiledgeX
...
-    public static IEnumerator InitalizeLocationService(int maxWait = 20, bool continuousLocationService = true)
+    public static IEnumerator InitalizeLocationService(int maxWait = 20, bool continuousLocationService = false)
     {
       // First, check if user has location service enabled
       if (!Input.location.isEnabledByUser)
       {
         Debug.Log("MobiledgeX: Location Service disabled by user.");
-        yield break;
-      }
+#if UNITY_IOS
```
franklin-huang-mobiledgex, line 37: thanks for the fix here

Runtime/Scripts/LocationService.cs
```diff
@@ -72,9 +80,26 @@ public static IEnumerator InitalizeLocationService(int maxWait = 20, bool contin
       }
     }
 
-    public IEnumerator Start()
+    public static IEnumerator AssureLocation()
+    {
+         if (Input.location.status == LocationServiceStatus.Initializing)
+         {
+             yield return new WaitUntil(() => Input.location.status == LocationServiceStatus.Running);
```
franklin-huang-mobiledgex, line 87: we should probably have a timeout for this also

Runtime/Scripts/LocationService.cs
ah1053: We call Location Once and its being saved


 Awake () > InitializeLocation () > Input.location.start() > location is saved in Input.location.lastData (which is getOnly)

next frame after we get the location , the Input.location.status is LocationServiceStatus.Stopped

Therefore a valid check for location is input.lastData value is not Input.location.status

Runtime/Scripts/LocationService.cs
ah1053: The timeout will be thrown in InitializeLocation Coroutine if it exceeds the 20 seconds limit

Runtime/Scripts/LocationService.cs
lgarner: Nit: spaces :)

Runtime/Scripts/LocationService.cs
```diff
@@ -72,9 +80,26 @@ public static IEnumerator InitalizeLocationService(int maxWait = 20, bool contin
       }
     }
 
-    public IEnumerator Start()
+    public static IEnumerator AssureLocation()
```
lgarner, line 83: Probably want EnsureLocation per dictionary definition.

PR #66, opened 2020-07-23
Example Code

ah1053:
**Example Code** 
1. WebSocket Example
2. Rest Example (UnityWebRequest - HttpClient)
3. Deleting Example.cs (used to have the two examples above)


Runtime/Scripts/ExampleRest.cs
```diff
@@ -0,0 +1,68 @@
...
+            }
+            catch (DmeDnsException)
+            {
+                mxi.UseWifiOnly(true);
+                await mxi.RegisterAndFindCloudlet();
+            }
+            mxi.GetAppPort(LProto.L_PROTO_HTTP);
+            string url = mxi.GetUrl("http");
+            Debug.Log("Rest URL is : " + url);
+
```
lgarner, line 37: // Tiny comment about: do something with the new edge server URL "here"

Runtime/Scripts/ExampleRest.cs
```diff
@@ -0,0 +1,68 @@
...
+            StartCoroutine(RestExample(url)); // You can use RestExampleHttpClient
```
lgarner, line 38: "You can instead use..."

Runtime/Scripts/ExampleRest.cs
ah1053: done

Runtime/Scripts/ExampleRest.cs
```diff
@@ -0,0 +1,68 @@
...
+            }
+            catch (DmeDnsException)
+            {
+                mxi.UseWifiOnly(true);
+                await mxi.RegisterAndFindCloudlet();
+            }
+            mxi.GetAppPort(LProto.L_PROTO_HTTP);
+            string url = mxi.GetUrl("http");
+            Debug.Log("Rest URL is : " + url); // Once you have your edge server url you can start communicating with your Edge sever deployed on MobiledgeX Console
+
```
brucearmstrong, line 37: nit: sever->server

PR #67, opened 2020-08-01
EDGECLOUD-3337: Unity: App Independent way to generate a Unique ID

lgarner:
Unity side of things. Android and IOS. Android has a separate implementation that is asynchronous via MobiledgeXAgent game object. It is not guaranteed it'll get a value before the rest of the app tries to use the value.

Requires C# SDK also updated to 2.1.3.


Runtime/Scripts/MobiledgeXAgent.cs
```diff
@@ -0,0 +1,162 @@
...
+    {
+      lock (this)
+      {
+        NotifyHashAvertisingIdCallers -= request;
+      }
+    }
+
+    IEnumerator GetHashedAdvertisingID()
+    {
+      string hashedHexAdId = null;
```
ah1053, line 126: Are you using hashedHexAdId somewhere ?

Runtime/Scripts/MobiledgeXAgent.cs
```diff
@@ -0,0 +1,162 @@
...
+
+    public const int DEFAULT_ADVERTISING_ID_TIMEOUT_MS = 3000;
+    public int HashedAdvertisingIdTimeoutMs { get; set; } = DEFAULT_ADVERTISING_ID_TIMEOUT_MS;
+
+    private bool hashAdvertisingIdRequestedPrivate = false;
+    private bool hashAdvertisingIdRequested
+    {
+      get
+      {
+        string tmp = null;
```
ah1053, line 48: tmp is not used ?!

Runtime/Scripts/MobiledgeXAgent.cs
```diff
@@ -0,0 +1,162 @@
...
+      yield return StartCoroutine(GetHashedAdvertisingID());
+    }
+
+    // Update is called once per frame
+    void Update()
+    {
+      if (hashAdvertisingIdRequested)
+      {
+        //Debug.Log("ZZZ: Starting AdID coroutine:");
+        StartCoroutine(GetHashedAdvertisingID());
```
ah1053, line 76: I don't know if you need to call **GetHashedAdvertisingID** in Update also
Although you have a lock its being called two times
Screen Shot 2020-08-03 at 11 27 44 AM

Runtime/Scripts/MobiledgeXIntegration.cs
```diff
@@ -93,6 +96,10 @@ public MobiledgeXIntegration()
             matchingEngine = new MatchingEngine(pIntegration.CarrierInfo, pIntegration.NetInterface, pIntegration.UniqueID);
             melMessaging = new MelMessaging(appName);
             matchingEngine.SetMelMessaging(melMessaging);
+
+            Debug.Log("ZZZ: Manufacturer: " + melMessaging.GetManufacturer());
+            Debug.Log("ZZZ: Device Type: " + pIntegration.UniqueID.GetUniqueIDType());
+            Debug.Log("ZZZ: Device Ad ID: " + pIntegration.UniqueID.GetUniqueID());
```
franklin-huang-mobiledgex, line 102: ZZZ?

lgarner: ^ Will fix, we had a change of plans...

Runtime/Scripts/MobiledgeXAgent.cs
lgarner: Impl now changed. Not needed currently

Runtime/Scripts/MobiledgeXAgent.cs
lgarner: Removed function.

Runtime/Scripts/MobiledgeXAgent.cs
lgarner: Removed. Stale.

lgarner:
Simplified for now, but largely the same. 2.1.3 C# SDK update required still.

PR #72, opened 2020-08-12
EDGECLOUD-3430: Unity SDK - MobiledgeX Integration script errors keeping the MobiledgeX package from being installed 

franklin-huang-mobiledgex:
1) Switched order of parameters for CreateUrl in C# SDK -> update in Unity SDK
2) Add Unity SDK support for Linux and Windows10

### DIFF
--- a/Runtime/Scripts/MobiledgeXIntegration.cs
+++ b/Runtime/Scripts/MobiledgeXIntegration.cs
@@ -292,7 +292,7 @@ namespace MobiledgeX
                 appPort = latestAppPort;
             }
 
-            return matchingEngine.CreateUrl(latestFindCloudletReply, appPort, port, l7Proto, path);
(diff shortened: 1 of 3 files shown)


Runtime/Scripts/MobiledgeXIntegration.cs
```diff
@@ -292,7 +292,7 @@ public string GetUrl(string l7Proto, AppPort appPort = null, int port = 0, strin
                 appPort = latestAppPort;
             }
 
-            return matchingEngine.CreateUrl(latestFindCloudletReply, appPort, port, l7Proto, path);
+            return matchingEngine.CreateUrl(latestFindCloudletReply, appPort, l7Proto, port, path);
```
lgarner, line 295: This swap already in the C# SDK binary?

franklin-huang-mobiledgex: yes. when i updated port validation

PR #79, opened 2020-09-01
Expose Location Parameter for Developers

ah1053:
https://mobiledgex.atlassian.net/browse/EDGECLOUD-3157

Location is used in **RegisterClient**  (isRoaming check) & **FindCloudlet**

Disable LocationService in the Inspector and use SetupLocation(Loc) before Register&FindCloudlet()
* no need to use LocationService.EnsureLocation if you will use SetupLocation()

Screen Shot 2020-08-31 at 6 12 36 PM

### DIFF
--- a/Runtime/Scripts/LocationService.cs
+++ b/Runtime/Scripts/LocationService.cs
@@ -30,9 +30,18 @@ namespace MobiledgeX
     {
         private static bool locationPermissionRejected = false;
 
+        [Tooltip("Disable LocationService flow if you will use SetupLocation() to supply your the location on your own")]
+        [SerializeField]
+        bool useLocationServiceFlow = true;
+
+        static Loc locationSupplied;
+
         void Awake()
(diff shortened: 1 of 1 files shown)


Runtime/Scripts/LocationService.cs
```diff
@@ -147,9 +156,20 @@ public IEnumerator LocationServiceFlow()
...
+        public static void SetupLocation(Loc location)
+        {
+            locationSupplied = location;
+        }
+
         // Retrieve the lastest location, without restarting locationService.
         public static Loc RetrieveLocation()
         {
+            if (Input.location.lastData.Equals(default(LocationInfo))){
+
```
franklin-huang-mobiledgex, line 168: nit extra line, same as line 170

ah1053:
Will proceed with changing location through setFallbackLocation as advised by @lgarner

PR #83, opened 2020-09-09
EDGECLOUD-3583: Remove unnecessary MEL Port check.

lgarner:
This check is no longer necessary for MEL, and in fact has all ports, not just one against the updated server. Android is unaffected.


brucearmstrong (approved):

ah1053 (approved):
⭐ 
Looks great @lgarner 
Can we also remove the recent change at MLO Unity SDK commit
Debug.LogError("Register OK")  
> MobiledgeXIntegration.cs line 118

It's being logged as an error when its not

LeonAdams (approved):
I am all for removing unnecessary code! Makes it easier to read

PR #84, opened 2020-09-11
Regions in Editor only

ah1053:
DME regions in Editor Only for testing
ComputerVision package EU DME
Screen Shot 2020-09-11 at 2 42 08 PM
As discussed with @franklin-huang-mobiledgex  for testing and developers anywhere would be able to to access our GPU servers that exist only in EU  (at least in the editor for now)


Editor/MobiledgeXEditorWindow.cs
```diff
@@ -68,6 +68,35 @@ private Texture2D MakeTex(int width, int height, Color col)
...
+                {
+                    case "EU":
+                        return EU_DME;
+                    case "KR":
+                        return KR_DME;
+                    case "JP":
+                        return JP_DME;
+                    case "US":
+                        return US_DME;
+                    case "Any":
```
franklin-huang-mobiledgex, line 89: Should we make this "Nearest" instead of "Any"?

Editor/MobiledgeXEditorWindow.cs
```diff
@@ -349,8 +384,9 @@ public async Task<bool> CheckCredentials()
             {
                 // Register and find cloudlet:
                 clog("Registering to DME ...", "");
-                checkResult = await integration.Register();
-                bool foundCloudlet = await integration.FindCloudlet();
+                checkResult = await integration.Register(Region == "" ? "" : Region, Region == "" ? (uint) 0 : DME_PORT);
```
franklin-huang-mobiledgex, line 387: dmeHost can just be Region. Don't think you need the ternary operator

Editor/MobiledgeXEditorWindow.cs
ah1053: Done

Editor/MobiledgeXEditorWindow.cs
ah1053: As discussed it if Nearest is chosen it will perform A regular RegisterAndFindCloudlet("",0)

franklin-huang-mobiledgex: This comment is about the first parameter. It can just be integration.Register(Region, Region == "" ? uint(0) : DME_PORT);

Editor/MobiledgeXEditorWindow.cs
```diff
@@ -68,6 +68,35 @@ private Texture2D MakeTex(int width, int height, Color col)
...
+                    default:
+                        return "";
+                }
+            }
+        }
+        const string EU_DME = "eu-mexdemo.dme.mobiledgex.net";
+        const string KR_DME = "kr-mexdemo.dme.mobiledgex.net";
+        const string US_DME = "us-mexdemo.dme.mobiledgex.net";
+        const string JP_DME = "jp-mexdemo.dme.mobiledgex.net";
+        const uint DME_PORT = 38001;
```
lgarner, line 99: If in namespace, use the MatchingEngine constant, one source of data only, if possible.

Editor/MobiledgeXEditorWindow.cs
lgarner: This is confusing to the eye, ternary ops should make things easier to understand at first glance, not harder.

How about you just put it can another line like at 273, and then use that?

Editor/MobiledgeXEditorWindow.cs
```diff
@@ -68,6 +68,35 @@ private Texture2D MakeTex(int width, int height, Color col)
...
+                        return EU_DME;
+                    case "KR":
+                        return KR_DME;
+                    case "JP":
+                        return JP_DME;
+                    case "US":
+                        return US_DME;
+                    case "Nearest":
+                    default:
+                        return "";
```
lgarner, line 91: How about pull default/WiFi from the matching engine constant itself?

Editor/MobiledgeXEditorWindow.cs
ah1053: Good catch, Thanks

Editor/MobiledgeXEditorWindow.cs
ah1053: What is the difference ?

franklin-huang-mobiledgex (approved):
Should be useful!

Editor/MobiledgeXEditorWindow.cs
ah1053: you have a point, I added wifi.dme it will make it the code look cleaner.

Editor/MobiledgeXEditorWindow.cs
```diff
@@ -68,6 +68,35 @@ private Texture2D MakeTex(int width, int height, Color col)
...
+                {
+                    case "EU":
+                        return EU_DME;
+                    case "KR":
+                        return KR_DME;
+                    case "JP":
+                        return JP_DME;
+                    case "US":
+                        return US_DME;
+                    case "Nearest":
```
franklin-huang-mobiledgex, line 89: no need for this case

brucearmstrong (approved):

Editor/MobiledgeXEditorWindow.cs
```diff
@@ -68,6 +68,35 @@ private Texture2D MakeTex(int width, int height, Color col)
...
+                    default:
+                        return WIFI_DME;
+                }
+            }
+        }
+        const string WIFI_DME = "wifi.dme.mobiledgex.net";
```
lgarner, line 95: Dup: public const string wifiOnlyDmeHost = wifiCarrier + "." + baseDmeHost; // but if it's not in the name space, then not terrible to duplicate among the ones below.

lgarner (approved):
👍

PR #89, opened 2020-09-22
CV Package Updated

ah1053:
This PR will not be merged to staging_2.1.y its for review only, it will be included in a the next PR as part of the CV Unity Package.
Main changes:
1. Connecting to tls CV server.
2. Image shrinking ratio, Please check the tool tip if there is any better phrasing or mistakes.

### DIFF
--- a/ComputerVision.cs
+++ b/ComputerVision.cs
@@ -1,0 +1,518 @@
+﻿
+using UnityEngine;
+using System;
+using System.Threading.Tasks;
+using System.Net;
+using System.Net.Sockets;
+using System.Linq;
+using MobiledgeX;
+using DistributedMatchEngine;
+using System.Collections;
+using UnityEngine.Networking;
+using System.Text;
+using System.Runtime.Serialization;
+using System.IO;
+using System.Runtime.Serialization.Json;
+
+[RequireComponent(typeof(MobiledgeX.LocationService))]
+[AddComponentMenu("MobiledgeX/ComputerVision")]
+public class ComputerVision : MonoBehaviour
+{
+    public ConnectionMode connectionMode;
+    public ServiceMode serviceMode = ServiceMode.ObjectDetection;
+
+    [Tooltip("The image will be resized before sending it to the server respecting the image aspect ratio. Image Shrinking Ratio is proportional to Detection Quality &" +
+        "inversely proportional to the Full Detection Process Latency. Image Shrinking Ratio is between (0.1 to 0.4)")]
+    [Range(0.1f, 0.4f)]
+    public float imageShrinkingRatio = 0.15f;
+
+    [Header("Face Detection")]
+    [Tooltip("Texture to be added on top of detected faces")]
+    public Texture faceRectTexture;
+
+    [Header("Object Detection")]
+    [Tooltip("For Object Detection, Texture to be added on top of detected objects with Confidence level higher than the confidence Threshold ")]
+    public Texture highConfidenceTexture;
+
+    [Tooltip("For Object Detection, Texture to be added on top of detected objects with Confidence level lower than the confidence Threshold ")]
+    public Texture lowConfidenceTexture;
+
+    public Font objectClassFont;
+    [Tooltip("Wether to show the confidence level next to the detected object or not, ex.(car 100%)")]
+    public bool showConfidenceLevel = true;
+
+    [Tooltip("For object detection, the font of the object class font")]
+    public int objectClassFontSize = 30;
+
+    [Tooltip("For object detection, color of the object class font if the confidence level is above or equal the confidence threshold")]
+    public Color highConfidenceFontColor = Color.green;
+
+    [Tooltip("For object detection, color of the object class font if the confidence level is below the confidence threshold")]
+    public Color lowConfidenceFontColor = Color.red;
+
+    public int confidenceThreshold = 1;
+
+    public enum ConnectionMode
+    {
+        Rest,
+        WebSocket
+    }
+
+    public enum ServiceMode
+    {
+        FaceDetection,
+        ObjectDetection
+    }
+
+    MobiledgeXWebSocketClient client;
+    MobiledgeXIntegration integration;
+    /// <summary>
+    /// Hide Detection rects before taking a screenshot
+    /// </summary>
+    static bool showGUI;
+    int[][] faceDetectionRects;
+    static @Object[] objectsDetected;
+
+
+    bool serviceAlreadyStarted
+    {
+        get
+        {
+            switch (serviceMode)
+            {
+                case ServiceMode.ObjectDetection:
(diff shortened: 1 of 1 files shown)


ComputerVision.cs
```diff
@@ -0,0 +1,518 @@
...
+using System.Runtime.Serialization.Json;
+
+[RequireComponent(typeof(MobiledgeX.LocationService))]
+[AddComponentMenu("MobiledgeX/ComputerVision")]
+public class ComputerVision : MonoBehaviour
+{
+    public ConnectionMode connectionMode;
+    public ServiceMode serviceMode = ServiceMode.ObjectDetection;
+
+    [Tooltip("The image will be resized before sending it to the server respecting the image aspect ratio. Image Shrinking Ratio is proportional to Detection Quality &" +
```
brucearmstrong, line 24: Nit: lack of space after ampersand

ComputerVision.cs
```diff
@@ -0,0 +1,518 @@
...
+
+    [Header("Object Detection")]
+    [Tooltip("For Object Detection, Texture to be added on top of detected objects with Confidence level higher than the confidence Threshold ")]
+    public Texture highConfidenceTexture;
+
+    [Tooltip("For Object Detection, Texture to be added on top of detected objects with Confidence level lower than the confidence Threshold ")]
+    public Texture lowConfidenceTexture;
+
+    public Font objectClassFont;
+    [Tooltip("Wether to show the confidence level next to the detected object or not, ex.(car 100%)")]
```
brucearmstrong, line 41: Nit: "Whether"

ComputerVision.cs
```diff
@@ -0,0 +1,518 @@
...
+    string urlSuffix
+    {
+        get
+        {
+            switch (serviceMode)
+            {
+                case ServiceMode.ObjectDetection:
+                    return "/object/detect/";
+                case ServiceMode.FaceDetection:
+                    return "/detector/detect/";
```
brucearmstrong, line 107: No Face Detection?

ComputerVision.cs
```diff
@@ -0,0 +1,518 @@
...
+                url = integration.GetUrl("wss");
+                return url;
+            case ConnectionMode.Rest:
+                appPort = integration.GetAppPort(LProto.L_PROTO_TCP);
+                url = integration.GetUrl("https");
+                return url;
+            default:
+                return "";
+        }
+    }
```
brucearmstrong, line 211: No http and ws for non-TLS app instances? Since we are controlling the backend deployments, and this app is specific to CV, I guess this is OK.

brucearmstrong (approved):
Nice, clean code. See the few comments.

ComputerVision.cs
```diff
@@ -0,0 +1,518 @@
...
+    MobiledgeXWebSocketClient client;
+    MobiledgeXIntegration integration;
+    /// <summary>
+    /// Hide Detection rects before taking a screenshot
+    /// </summary>
+    static bool showGUI;
+    int[][] faceDetectionRects;
+    static @Object[] objectsDetected;
+
+
```
franklin-huang-mobiledgex, line 76: nit extra line

ComputerVision.cs
```diff
@@ -0,0 +1,518 @@
...
+    #region Monobehaviour callbacks
+    IEnumerator Start()
+    {
+#if UNITY_EDITOR
+        GetEdgeConnection();
+        yield break;
+#endif
+        if (Application.internetReachability == NetworkReachability.NotReachable)
```
franklin-huang-mobiledgex, line 131: hmm. Is this a Unity property? Why this over HasCellular check or IsEdgeEnabled?

ComputerVision.cs
```diff
@@ -0,0 +1,518 @@
...
+            return;
+        }
+        var cqueue = client.receiveQueue;
+        string msg;
+        while (cqueue.TryPeek(out msg))
+        {
+            cqueue.TryDequeue(out msg);
+            HandleServerRespone(msg);
+        }
+
```
franklin-huang-mobiledgex, line 166: nit extra line

ComputerVision.cs
```diff
@@ -0,0 +1,518 @@
...
+    #endregion
+
+    async Task GetEdgeConnection()
+    {
+        integration = new MobiledgeXIntegration();
+        integration.UseWifiOnly(true);
+        
+        try
+        {
+            bool cloudletFound = await integration.RegisterAndFindCloudlet("eu-mexdemo.dme.mobiledgex.net", 38001);
```
franklin-huang-mobiledgex, line 177: hardcode?

ComputerVision.cs
```diff
@@ -0,0 +1,518 @@
...
+        await client.Connect(uri);
+        wsStarted = true;
+    }
+
+    public void SendtoServer(byte[] imgBinary)
+    {
+        client.Send(imgBinary);
+    }
+
+    public IEnumerator SendImageToServer(byte[] imgBinary, string url)
```
franklin-huang-mobiledgex, line 232: We do have a GetHttpClient function. But if this is more dependable, seems fine

franklin-huang-mobiledgex (approved):
Looks pretty good!

brucearmstrong:
Thinking about this some more, you do actually need to support non-TLS connections, because TLS is not possible on instances deployed to GCP cloudlets, which we do use in the US.

ComputerVision.cs
lgarner: Yes, wonder what the test is for this.

ComputerVision.cs
ah1053: As discussed in a different PR, In the next release we will have the SDK regions in the SDK and the CV component will pull from those values.
However we are connecting only to EU region since in this region is the only one who have GPU enabled Cloudlets.

ComputerVision.cs
```diff
@@ -0,0 +1,518 @@
...
+{
+    [DataMember]
+    public int[] rect;
+    [DataMember]
+    public string @class;
+    [DataMember]
+    public float confidence;
+}
+
+[DataContract]
```
lgarner, line 470: We may want to expose something this generic in the SDK itself.

ComputerVision.cs
ah1053: Good Point, After long time debugging, I think the best way for WebRequests in Unity is Coroutines, it doesn't block the main thread and have better performance.

ComputerVision.cs
ah1053: This will be super helpful for Indie Developers (90% of Unity Developer)

ComputerVision.cs
ah1053: It's  a simple check for Internet Connection
https://docs.unity3d.com/ScriptReference/NetworkReachability.html

PR #94, opened 2020-10-05
MobiledgeXUDPClient EDGECLOUD-3158

ah1053:
Usage example 
``` c-sharp
MobiledgeXUDPClient udpClient =   new MobiledgeXUDPClient(integration.GetHost(),
 integration.GetAppPort(LProto.L_PROTO_UDP).public_port,
 udpReceivePort);

udpClient.Send("msg") or send a byte array

void Update(){
    if (udpClient == null)
            {
                return;
            }
            //udp receive queue
            byte[] udpMsg;
            var udp_queue = udpClient.receiveQueue;
            while (udp_queue.TryPeek(out udpMsg))
            {
                udp_queue.TryDequeue(out udpMsg);
                HandleUDPMessage(udpMsg);
            }
        }

```
According to [RFC 1122](https://tools.ietf.org/html/rfc1122) Max deliverable packet size is 576 

(MAX PAYLOAD = 508 = 576 - 60 IP Header - 8 UDP Header)
Screen Shot 2020-10-05 at 2 50 52 PM

### DIFF
--- a/Runtime/Scripts/MobiledgeXUDPClient.cs
+++ b/Runtime/Scripts/MobiledgeXUDPClient.cs
@@ -1,0 +1,144 @@
+﻿/**
+ * Copyright 2018-2020 MobiledgeX, Inc. All rights and licenses reserved.
+ * MobiledgeX, Inc. 156 2nd Street #408, San Francisco, CA 94105
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+using System;
+using System.Collections.Concurrent;
+using System.Net;
+using System.Net.Sockets;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using UnityEngine;
+using System.Linq;
+namespace MobiledgeX
+{
+    // MobiledgeXUDPClient is a UDP Client Implementation offered with MobiledgeX Unity Package
+    // MobiledgeXUDPClient concurrency model supports the use of a single queue for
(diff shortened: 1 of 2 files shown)


Runtime/Scripts/MobiledgeXUDPClient.cs
```diff
@@ -0,0 +1,144 @@
...
+        {
+            ArraySegment<byte> msg;
+            Debug.Log("RunSend entered.");
+            while (run)
+            {
+                while (!sendQueue.IsCompleted)
+                {
+                    msg = sendQueue.Take();
+                    long count = sendQueue.Count;
+                    Debug.Log("Dequeued this message to send: " + msg + ", queueSize: " + count);
```
lgarner, line 112: Config for verbose mode?

Runtime/Scripts/MobiledgeXUDPClient.cs
```diff
@@ -0,0 +1,144 @@
...
+            run = true;
+        }
+
+        public void Send(string message)
+        {
+
+            byte[] buffer = encoder.GetBytes(message);
+            if (buffer.Length > MAXPAYLOADSIZE)
+            {
+                Debug.LogError("Max UDP payload size is 508 bytes, trying slicing your message to suit the max payload size");
```
lgarner, line 84: " + MAXPAYLOADSIZE + ", for consistency between const and text.

Runtime/Scripts/MobiledgeXUDPClient.cs
ah1053: Thats a good suggestion we should do that for the whole SDK , will open a ticket for that

brucearmstrong (approved):

Runtime/Scripts/MobiledgeXUDPClient.cs
```diff
@@ -0,0 +1,144 @@
...
+        public void Connect()
+        {
+            run = true;
+        }
+
+        public void Send(string message)
+        {
+
+            byte[] buffer = encoder.GetBytes(message);
+            if (buffer.Length > MAXPAYLOADSIZE)
```
franklin-huang-mobiledgex, line 82: hmm were you running into issues with messages that were too large? I don't think you need to check max datagram size, since I'm assuming the C# UDP implementation handles the RFC specs and will split up the message into separate datagrams

franklin-huang-mobiledgex (approved):

PR #95, opened 2020-10-08
Regions In Editor & useSelectedRegionInProduction EDGECLOUD-3760

ah1053:
(Regions in Editor)
The changes discussed on Slack:
In MobiledgeX Editor Window a developer can select a region , Register() , FindCloudlet() will be executed using the selected region DME **only in the UnityEditor**.
If a developer wants to override the DME HOST for production for non sim card device (ex. Magic Leap)
```c-sharp
integration.useSelectedRegionInProduction = true;
integration.RegisterAndFindCloudlet();
```
@lgarner  this will affect the MagicLeap Demo so you will need something like this  https://github.com/mobiledgex/unity-samples/pull/5

Added check to prevent region corruption in the inspector

### DIFF
--- a/Runtime/Scripts/MobiledgeXIntegration.cs
+++ b/Runtime/Scripts/MobiledgeXIntegration.cs
@@ -134,16 +134,16 @@ namespace MobiledgeX
         /// RegisterClientException and FindCloudletException will give more details on reason for failure
         /// </summary>
         /// <returns>bool Task</returns>
-        public async Task<bool> RegisterAndFindCloudlet()
+        public async Task<bool> RegisterAndFindCloudlet(string dmeHost = "", uint dmePort = 0)
         {
-            bool registered = await Register();
+            bool registered = await Register(dmeHost, dmePort);
             if (!registered)
             {
                 Debug.LogError("Register Failed!");
                 return false;
             }
             Debug.Log("Register OK!");
-            bool found = await FindCloudlet();
+            bool found = await FindCloudlet(dmeHost, dmePort);
             if (!found)
             {
               Debug.LogError("FindCloudlet Failed!");
--- a/Runtime/Scripts/MobiledgeXIntegration.cs
+++ b/Runtime/Scripts/MobiledgeXIntegration.cs
@@ -83,6 +83,33 @@ namespace MobiledgeX
         CarrierInfoClass carrierInfoClass = new CarrierInfoClass(); // used for IsRoaming check
         MelMessaging melMessaging;
 
+        string region
+        {
+            get
+            {
+                switch (settings.region)
+                {
+                    case "EU":
+                        return EU_DME;
+                    case "KR":
+                        return KR_DME;
+                    case "JP":
+                        return JP_DME;
+                    case "US":
+                        return US_DME;
+                    case "Nearest":
+                    default:
+                        return WIFI_DME;
+                }
+            }
+        }
+
+        const string WIFI_DME = "wifi.dme.mobiledgex.net";
+        const string EU_DME = "eu-mexdemo.dme.mobiledgex.net";
+        const string KR_DME = "kr-mexdemo.dme.mobiledgex.net";
+        const string US_DME = "us-mexdemo.dme.mobiledgex.net";
+        const string JP_DME = "jp-mexdemo.dme.mobiledgex.net";
+
         /// <summary>
         /// Constructor for MobiledgeXIntegration. This class has functions that wrap DistributedMatchEngine functions for ease of use
         /// </summary>
@@ -107,16 +134,16 @@ namespace MobiledgeX
         /// RegisterClientException and FindCloudletException will give more details on reason for failure
         /// </summary>
         /// <returns>bool Task</returns>
-        public async Task<bool> RegisterAndFindCloudlet(string dmeHost = null, uint dmePort = 0)
+        public async Task<bool> RegisterAndFindCloudlet()
         {
-            bool registered = await Register(dmeHost, dmePort);
+            bool registered = await Register();
             if (!registered)
             {
                 Debug.LogError("Register Failed!");
                 return false;
             }
             Debug.Log("Register OK!");
-            bool found = await FindCloudlet(dmeHost, dmePort);
+            bool found = await FindCloudlet();
             if (!found)
             {
               Debug.LogError("FindCloudlet Failed!");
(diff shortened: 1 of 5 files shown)


brucearmstrong (approved):

Runtime/Scripts/MobiledgeXIntegrationHelper.cs
```diff
@@ -61,12 +61,20 @@ public partial class MobiledgeXIntegration
         /// </summary>
         public bool useFallbackLocation = false;
 
+        /// <summary>
+        /// You don't need this option in UnityEditor by default the region used will be the region selected in MobiledgeX Editor Window
+        /// Set to true to use the Region selected in MobiledgeX Editor Window in production
```
lgarner, line 66: . It's

Runtime/Scripts/MobiledgeXIntegrationHelper.cs
```diff
@@ -61,12 +61,20 @@ public partial class MobiledgeXIntegration
...
+        /// its not recommended to use this option in production since the SDK will automatically select the best region
```
lgarner, line 67: region. You

Runtime/Scripts/MobiledgeXIntegrationHelper.cs
```diff
@@ -82,44 +90,50 @@ public async Task<bool> Register(string dmeHost = null, uint dmePort = 0)
...
+#else
+                    Debug.Log("MobiledgeX: Doing Register Client, with req: " + req);
                     reply = await matchingEngine.RegisterClient(req);
+#endif
                 }
                 else
                 {
-                    Debug.Log("Doing Register Client with DME: " + dmeHost + ", p: " + dmePort + " with req: " + req);
-                    reply = await matchingEngine.RegisterClient(dmeHost, dmePort, req);
+                    Debug.Log("MobiledgeX: Doing Register Client with DME: " + region + ", p: " + MatchingEngine.defaultDmeRestPort + " with req: " + req);
```
lgarner, line 113: Region is not mcc-mnc, and the parameters have removed dme and port to allow SIM usage, potentially breaking some devices.

ah1053: I have (Editor Only) comment next to the region
Screen Shot 2020-10-14 at 11 37 36 PM

lgarner (approved):
Ok. I'll need to update some downstream code. Should be fine.

brucearmstrong (approved):

Runtime/Scripts/MobiledgeXIntegrationHelper.cs
```diff
@@ -82,44 +90,50 @@ public async Task<bool> Register(string dmeHost = null, uint dmePort = 0)
...
             try
             {
-                if (dmeHost == null || dmePort == 0)
+                if (!useSelectedRegionInProduction)
                 {
-                    Debug.Log("Doing Register Client, with req: " + req);
+#if UNITY_EDITOR
+                        Debug.Log("MobiledgeX: Doing Register Client with DME: " + region + ", p: " + MatchingEngine.defaultDmeRestPort + " with req: " + req);
+                        Debug.LogWarning("MobiledgeX: Region Selection will work only in UnityEditor not on Mobile Devices");
+                        reply = await matchingEngine.RegisterClient(region, MatchingEngine.defaultDmeRestPort, req);
```
franklin-huang-mobiledgex, line 105: Nit extra tabs

Runtime/Scripts/MobiledgeXIntegration.cs
```diff
@@ -107,16 +134,16 @@ public MobiledgeXIntegration(CarrierInfo carrierInfo = null, NetInterface netInt
         /// RegisterClientException and FindCloudletException will give more details on reason for failure
         /// </summary>
         /// <returns>bool Task</returns>
-        public async Task<bool> RegisterAndFindCloudlet(string dmeHost = null, uint dmePort = 0)
+        public async Task<bool> RegisterAndFindCloudlet()
         {
-            bool registered = await Register(dmeHost, dmePort);
+            bool registered = await Register();
```
franklin-huang-mobiledgex, line 139: hmm. Didn't we want the dmeHost and dmePort for QA to be able to test?

lgarner: Actually, yes, they're optional parameters at the C# SDK level for testing reasons. If the overrides are expansive enough, or "user specified" override, that can be enough, though I tend to just want optional parameters that are stateless, rather than contextual, which implies state management.

Runtime/Scripts/MobiledgeXIntegrationHelper.cs
ah1053: Done

Runtime/Scripts/MobiledgeXIntegration.cs
```diff
@@ -83,6 +83,33 @@ public partial class MobiledgeXIntegration
...
+                        return JP_DME;
+                    case "US":
+                        return US_DME;
+                    case "Nearest":
+                    default:
+                        return WIFI_DME;
+                }
+            }
+        }
+
```
lgarner, line 106: This set might be sufficient for QA, but they may want to use their own staging server as well.

franklin-huang-mobiledgex: I don't think we want the QA dme information to be readily available. I think the host and port overrides are a better solution

lgarner: ^ True, that's what I added back. The alternate is a blank field to type in.

Runtime/Scripts/MobiledgeXIntegration.cs
```diff
@@ -107,7 +134,7 @@ public MobiledgeXIntegration(CarrierInfo carrierInfo = null, NetInterface netInt
         /// RegisterClientException and FindCloudletException will give more details on reason for failure
         /// </summary>
         /// <returns>bool Task</returns>
-        public async Task<bool> RegisterAndFindCloudlet(string dmeHost = null, uint dmePort = 0)
+        public async Task<bool> RegisterAndFindCloudlet(string dmeHost = "", uint dmePort = 0)
```
franklin-huang-mobiledgex, line 137: Think you can leave this as null. Default value of string anyways

franklin-huang-mobiledgex (approved):
Looks good. I suggest leaving the default value of dmeHost as null

PR #99, opened 2020-11-03
Fallback Location from IP Address EDGECLOUD-3750

ah1053:
**Currently**, We use a hard coded location as the default for the Fallback location (SanJose)
**New approach** is
 Check if Fallback Location is empty ,if so get location from the IP Address using this [api](https://freegeoip.app/) , In case of failure or any exception thrown use SanJose as the default  fallback location.
 >If the developer already used SetFallbackLocation() this request won't be used

See full description of the feature [here](https://mobiledgex.atlassian.net/browse/EDGECLOUD-3750)

### DIFF
--- a/Runtime/Scripts/MobiledgeXIntegrationConfig.cs
+++ b/Runtime/Scripts/MobiledgeXIntegrationConfig.cs
@@ -19,6 +19,11 @@ using System;
 using UnityEngine;
 using DistributedMatchEngine;
 using System.Threading.Tasks;
+using System.Runtime.Serialization;
+using System.Runtime.Serialization.Json;
+using System.IO;
+using System.Text;
+using System.Net.Http;
 
 /*
 * Utility functions to help developer configure their MobiledgeXIntegration object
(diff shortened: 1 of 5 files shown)


brucearmstrong (approved):
Great idea, and looks like a straightforward implementation.

Runtime/Scripts/MobiledgeXIntegrationConfig.cs
```diff
@@ -74,5 +79,50 @@ public void SetFallbackLocation(double longitude, double latitude)
...
+        {
+            HttpClient httpClient = new HttpClient();
+            try
+            {
+                HttpResponseMessage response = await httpClient.GetAsync("https://freegeoip.app/json/").ConfigureAwait(false);
+                string responseBodyStr = response.Content.ReadAsStringAsync().Result;
+                LocationFromIPAddress location = Messaging<LocationFromIPAddress>.Deserialize(responseBodyStr);
+                return location;
+            }
+            catch (Exception)
```
franklin-huang-mobiledgex, line 93: can we log the exception?

ah1053: Why ?

Runtime/Scripts/MobiledgeXIntegrationConfig.cs
```diff
@@ -74,5 +79,50 @@ public void SetFallbackLocation(double longitude, double latitude)
...
+        [DataContract]
+        public class LocationFromIPAddress
+        {
+            [DataMember]
+            public float longitude;
+            [DataMember]
+            public float latitude;
+        }
+
+        public static class Messaging<T>
```
franklin-huang-mobiledgex, line 112: Do we have this class in c-sharp sdk? Seems useful and wouldn't need to duplicate code

franklin-huang-mobiledgex (approved):
Looks good. Good idea!

Runtime/Scripts/MobiledgeXIntegrationConfig.cs
franklin-huang-mobiledgex: Might be useful when testing. But also not that important

Runtime/Scripts/MobiledgeXIntegrationConfig.cs
ah1053: I tried to search for it, It's only in the PingPongGame.
@lgarner 
Do we have this Messaging class somewhere in the C# SDK?

lgarner: iOS has a custom deserializer, but this is useful code otherwise were it not for that.

lgarner (approved):
extra space on return, but it'll also be overwritten soon...

PR #100, opened 2020-11-04
EDGECLOUD-3907: Objective C Unity Extra Device Info

franklin-huang-mobiledgex:
Update Objective C plugin with functions that grab device info (char* _getManufacturerCode(), char* _getDeviceSoftwareVersion(), char* _getDeviceModel(), char* _getOperatingSystem()). 

I use these functions in the DeviceInfoClass:GetDeviceInfo function to populate the device info dictionary.

(Note: ISO code from gps location requires async code, so I added an async GetDeviceInfo() function in MobiledgeXIntegration that adds "NetworkCountryIso" to the device info dictionary after grabbing it from DeviceInfoClass)

### DIFF
--- a/Runtime/Scripts/ExampleRest.cs
+++ b/Runtime/Scripts/ExampleRest.cs
@@ -7,8 +7,9 @@ using DistributedMatchEngine;
 using System.Threading.Tasks;
 using System.Net.Http;
 using System;
+using System.Collections.Generic;
 
-    [RequireComponent(typeof(MobiledgeX.LocationService))]
+[RequireComponent(typeof(MobiledgeX.LocationService))]
     public class ExampleRest : MonoBehaviour
     {
         MobiledgeXIntegration mxi;
@@ -22,6 +23,11 @@ using System;
         async void GetEdgeConnection()
         {
             mxi = new MobiledgeXIntegration();
+            var deviceInfo = await mxi.GetDeviceInfo();
+            foreach (KeyValuePair<string, string> pair in deviceInfo)
+            {
+                Debug.Log("Key is " + pair.Key + ", and Value is " + pair.Value);
+            }
             try
             {
                 await mxi.RegisterAndFindCloudlet();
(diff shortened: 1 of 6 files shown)


Runtime/Scripts/MobiledgeXIntegrationHelper.cs
```diff
@@ -316,6 +317,19 @@ public async Task<bool> IsRoaming()
         }
 #endif
 
+        public async Task<Dictionary<string, string>> GetDeviceInfo()
```
lgarner, line 320: Okay, extra layer, and Caller is async already.

brucearmstrong (approved):

Runtime/Scripts/ExampleRest.cs
```diff
@@ -22,6 +23,11 @@ IEnumerator Start()
         async void GetEdgeConnection()
         {
             mxi = new MobiledgeXIntegration();
+            var deviceInfo = await mxi.GetDeviceInfo();
+            foreach (KeyValuePair<string, string> pair in deviceInfo)
+            {
+                Debug.Log("DeviceInfo: Key is " + pair.Key + ", and Value is " + pair.Value);
+            }
```
ah1053, line 30: I think you might forgot this here when you were debugging, but I don't think this should be here ?

Resources/Examples/EdgeMultiplay.unitypackage.meta
```diff
@@ -0,0 +1,7 @@
+fileFormatVersion: 2
+guid: 9e9fc787c1d4f4beb98f96bc62e53166
+DefaultImporter:
+  externalObjects: {}
+  userData: 
+  assetBundleName: 
+  assetBundleVariant:
```
ah1053, line 7: Thanks

Runtime/Scripts/MobiledgeXIntegrationHelper.cs
franklin-huang-mobiledgex: Im guessing this will get sent out in RegisterClient

Runtime/Scripts/ExampleRest.cs
franklin-huang-mobiledgex: ya i can remove

ah1053:
I guess will have to hold on this update, till we get it to work on iOS 14.
More info in our slack conversations.

PR #106, opened 2020-11-27
UDP Example & Updated Readme

ah1053:
ExampleUDP.cs similar to ExampleRest.cs & ExampleWebSocket.cs

Updated Readme for UDPClient

### DIFF
--- a/Runtime/Scripts/ExampleUDP.cs
+++ b/Runtime/Scripts/ExampleUDP.cs
@@ -22,16 +22,7 @@ public class ExampleUDP : MonoBehaviour
     async void GetEdgeConnection()
     {
         mxi = new MobiledgeXIntegration();
-        try
-        {
-            await mxi.RegisterAndFindCloudlet();
-        }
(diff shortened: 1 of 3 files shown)


franklin-huang-mobiledgex (approved):
Looks good!

Runtime/Scripts/ExampleUDP.cs
```diff
@@ -0,0 +1,72 @@
...
+    async void GetEdgeConnection()
+    {
+        mxi = new MobiledgeXIntegration();
+        try
+        {
+            await mxi.RegisterAndFindCloudlet();
+        }
+        catch (DmeDnsException dde)
+        {
+            Debug.Log("Dme dns Exception: " + dde.Message);
```
lgarner, line 31: There a IsEdgeEnabled call to ensure the app has some kind of current available network status for the example?

ah1053: No

PR #115, opened 2021-02-03
Staging grpc v2.5

lgarner:
Initial GRPC version with events.


brucearmstrong (approved):
Looks pretty straightforward.

Runtime/Scripts/MobiledgeXIntegrationHelper.cs
```diff
@@ -227,16 +228,21 @@ public async Task<bool> FindCloudlet(string dmeHost = null, uint dmePort = 0)
...
             latestFindCloudletReply = reply;
-            latestAppPortList = reply.ports;
-            return reply.status == FindCloudletReply.FindStatus.FIND_FOUND;
+            int porti = 0;
+            latestAppPortList = new AppPort[reply.Ports.Count];
+            foreach (var aport in reply.Ports)
+            {
+               latestAppPortList[porti++] = aport;
+            }
+            return reply.Status == FindCloudletReply.Types.FindStatus.FindFound;
```
lgarner, line 245: BTW, we really ought to return the actual enum back in REST and GRPC, it's useful, versus the boolean.

Runtime/Scripts/MobiledgeXIntegration.cs
```diff
@@ -77,7 +77,7 @@ public partial class MobiledgeXIntegration
         bool latestVerifyLocationStatus = false; // Whether the most recent verifyLocation call was successful
         FindCloudletMode mode = FindCloudletMode.PROXIMITY; // FindCloudlet mode
         AppPort latestAppPort = null;
-        AppPort[] latestAppPortList = null;
+        AppPort [] latestAppPortList = null;
```
franklin-huang-mobiledgex, line 80: nit extra space

Runtime/Scripts/MobiledgeXIntegrationHelper.cs
franklin-huang-mobiledgex: Why? We do return in c#, but i'm guessing unity devs only care about if the findcloudlet was successful or not. Plus, returning FIND_NOTFOUND or FIND_UNKNOWN isn't very helpful in this case

franklin-huang-mobiledgex (approved):

Runtime/Scripts/MobiledgeXIntegrationHelper.cs
lgarner: It's not a guarantee there will never have new states (right now, it is mostly just the 2 states, Unknown shouldn't happen outside a Unity AOT reflection error (link.xml, etc), for example), and a switch statement isn't much more difficult to handle.

PR #116, opened 2021-02-03
EDGECLOUD-4300 Location Service Exception thrown when user rejects location permission

ah1053:
Location Exception thrown when user rejects location permission as discussed on Slack

### DIFF
--- a/Runtime/Scripts/ExampleRest.cs
+++ b/Runtime/Scripts/ExampleRest.cs
@@ -1,4 +1,21 @@
-﻿using System.Collections;
+﻿/**
+ * Copyright 2018-2021 MobiledgeX, Inc. All rights and licenses reserved.
+ * MobiledgeX, Inc. 156 2nd Street #408, San Francisco, CA 94105
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+using System.Collections;
 using UnityEngine;
 using UnityEngine.Networking; // for Unity Web Request
 
@@ -26,17 +43,24 @@ using System.Collections.Generic;
             try
             {
                 await mxi.RegisterAndFindCloudlet();
-            }
+            }           
             catch (DmeDnsException dde)
             {
                 Debug.Log("Dme dns Exception: " + dde.Message);
                 mxi.UseWifiOnly(true);
                 await mxi.RegisterAndFindCloudlet();
             }
+            // LocationException is thrown if app user rejected location permission
+            catch (LocationException locException)
+            {
+                print("Location Exception: " + locException.Message);
+                mxi.useFallbackLocation = true;
+                mxi.SetFallbackLocation(-122.4194, 37.7749); //Example only (SF location), You can use:  MobiledgeXIntegration.LocationFromIPAddress location = await MobiledgeXIntegration.GetLocationFromIP();
+                await mxi.RegisterAndFindCloudlet();
+            }
             mxi.GetAppPort(LProto.L_PROTO_TCP);
             string url = mxi.GetUrl("http");
             Debug.Log("Rest URL is : " + url); // Once you have your edge server url you can start communicating with your Edge server deployed on MobiledgeX Console
-
             StartCoroutine(RestExample(url)); //using UnityWebRequest
             //await RestExampleHttpClient(url); // You can instead use HttpClient
         }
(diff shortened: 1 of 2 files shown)


Runtime/Scripts/LocationService.cs
```diff
@@ -99,27 +97,39 @@ public static IEnumerator InitalizeLocationService(int maxWait = 20, bool contin
 
         /// <summary>
         /// EnsureLocation Confirm that user location is valid, user location is essential for MobiledgeX services
-        /// If Location permission is denied by User an exception will be thrown (for Android) in LocationServiceFlow()
-        ///                                                                      (for iOS) in InitalizeLocationService()
+        /// If Location permission is denied by User an exception will be thrown once RetreiveLocation() is called
```
brucearmstrong, line 100: Nit: Retreive -> Retrieve

brucearmstrong (approved):

Runtime/Scripts/LocationService.cs
lgarner: Actual function looks good.

Runtime/Scripts/ExampleRest.cs
```diff
@@ -26,17 +43,24 @@ async void GetEdgeConnection()
             try
             {
                 await mxi.RegisterAndFindCloudlet();
-            }
+            }
```
lgarner, line 46: Nit.

Runtime/Scripts/LocationService.cs
```diff
@@ -68,24 +64,22 @@ public static IEnumerator InitalizeLocationService(int maxWait = 20, bool contin
...
-#if UNITY_EDITOR
-                    Debug.LogWarning("MobiledgeX: Location Service disabled in UnityEditor");
-#else
-                    throw new Exception("MobiledgeX: Location Service disabled by user.");
-#endif
+                    locationPermissionRejected = true;
                 }
+#endif
                 // Access granted and location value could be retrieved
                 //Debug.Log("MobiledgeX: Location Service has location: " + Input.location.lastData.latitude + " " + Input.location.lastData.longitude + " " + Input.location.lastData.altitude + " " + Input.location.lastData.horizontalAccuracy + " " + Input.location.lastData.timestamp);
```
franklin-huang-mobiledgex, line 84: can you use your new Logger for this message?

Runtime/Scripts/LocationService.cs
```diff
@@ -28,8 +28,7 @@ namespace MobiledgeX
     [AddComponentMenu("MobiledgeX/LocationService")]
     public class LocationService : MonoBehaviour
     {
-        private static bool locationPermissionRejected = false;
-
+        public static bool locationPermissionRejected;
```
franklin-huang-mobiledgex, line 31: nit new line after the static var

Runtime/Scripts/LocationService.cs
```diff
@@ -40,18 +39,15 @@ public static IEnumerator InitalizeLocationService(int maxWait = 20, bool contin
...
                 //the CoreLocation framework prompts the user with a confirmation panel
                 //asking whether location services should be reenabled.
                 //The user can enable or disable location services altogether from the Settings application
                 //by toggling the switch in Settings>General>LocationServices.
                 //https://docs.unity3d.com/ScriptReference/LocationService-isEnabledByUser.html
-#elif UNITY_EDITOR
-                Debug.LogWarning("MobiledgeX: Location Service disabled in UnityEditor");
 #else
-                throw new Exception("MobiledgeX: Location Service disabled by user."); // 
+                locationPermissionRejected = true;
```
franklin-huang-mobiledgex, line 50: I think each time you set locationPermissionRejected to true in this function and in LocationServiceFlow(), we should Input.location.Stop(); and then return, right?

Runtime/Scripts/LocationService.cs
```diff
@@ -99,27 +93,39 @@ public static IEnumerator InitalizeLocationService(int maxWait = 20, bool contin
...
+
 #if UNITY_EDITOR
             yield return null;
 #else
             if (Input.location.status == LocationServiceStatus.Initializing)
             {
-                yield return new WaitUntil(() => (Input.location.status == LocationServiceStatus.Running));
+                yield return new WaitUntil(() => (Input.location.status != LocationServiceStatus.Initializing));
+            }
+            if(locationPermissionRejected == true || Input.location.status == LocationServiceStatus.Failed || !Input.location.isEnabledByUser)
```
franklin-huang-mobiledgex, line 112: nit: space between "if" and "("

franklin-huang-mobiledgex: Also, shouldn't the locationPermissionRejected check happen before we check the Input.location.status?

Runtime/Scripts/LocationService.cs
```diff
@@ -141,7 +146,12 @@ public IEnumerator LocationServiceFlow()
...
-            else //iOS - Permission Request are enabled once the application request location info
+            //iOS - Permission Request  once the application request location info
+            if (Application.platform == RuntimePlatform.IPhonePlayer)
+            {
+                yield return StartCoroutine(InitalizeLocationService());
+            }
+            else
             {
                 yield return StartCoroutine(InitalizeLocationService());
             }
```
franklin-huang-mobiledgex, line 157: Since it's the same code in both blocks, you can just add a single else block after the `if (Application.platform == RuntimePlatform.Android)`. Or if you think it may be different in the future, you can use a switch and falling through to the default

franklin-huang-mobiledgex (commented):
Looks good! See below comments about use of locationPermissionRejected. I think our location services should basically terminate if permissions are rejected

PR #117, opened 2021-02-04
Capture SDK Version only on DeviceInfo if the device  have Oculus in XR Settings EDGECLOUD-4335

ah1053:
Capture SDK Version only on DeviceInfoIntegration if the device is Oculus, GetTelephonyManager() cause multiple exceptions to be thrown on XR Devices that doesn't have Sim Card.

### DIFF
--- a/Runtime/Scripts/DeviceInfoIntegration.cs
+++ b/Runtime/Scripts/DeviceInfoIntegration.cs
@@ -64,11 +64,10 @@ namespace MobiledgeX
       CarrierInfoClass carrierInfo = new CarrierInfoClass();
       Dictionary<string, string> map;
       int sdk_int = carrierInfo.getAndroidSDKVers();
-      if (UnityEngine.XR.XRSettings.isDeviceActive && UnityEngine.XR.XRSettings.loadedDeviceName.Contains("oculus"))
+      if (UnityEngine.XR.XRSettings.enabled)
       {
-           map = new Dictionary<string, string>();
-           map["Build.VERSION.SDK_INT"] = sdk_int.ToString();
-           return map;
+          map["Build.VERSION.SDK_INT"] = sdk_int.ToString();
+          return map;
       }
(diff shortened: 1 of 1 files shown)


Runtime/Scripts/DeviceInfoIntegration.cs
```diff
@@ -45,20 +62,22 @@ enum NetworkDataType
     public Dictionary<string, string> GetDeviceInfo()
     {
       CarrierInfoClass carrierInfo = new CarrierInfoClass();
-      AndroidJavaObject telephonyManager = carrierInfo.GetTelephonyManager();
       Dictionary<string, string> map;
-
+      int sdk_int = carrierInfo.getAndroidSDKVers();
+      if (UnityEngine.XR.XRSettings.isDeviceActive && UnityEngine.XR.XRSettings.loadedDeviceName.Contains("oculus"))
```
lgarner, line 67: Ok. Once we have a better idea what DeviceInfo needs to be, this will be good enough. No telephony layer to get anything from at all.

Runtime/Scripts/DeviceInfoIntegration.cs
```diff
@@ -64,11 +64,10 @@ enum NetworkDataType
       CarrierInfoClass carrierInfo = new CarrierInfoClass();
       Dictionary<string, string> map;
       int sdk_int = carrierInfo.getAndroidSDKVers();
-      if (UnityEngine.XR.XRSettings.isDeviceActive && UnityEngine.XR.XRSettings.loadedDeviceName.Contains("oculus"))
+      if (UnityEngine.XR.XRSettings.enabled)
```
lgarner, line 67: Is XR never associated with Android Phones and iOS?

ah1053: Fixed to check for Oculus only

PR #119, opened 2021-02-05
MobiledgeXLogger

ah1053:
EDGECLOUD-4382 [MobiledgeXLogger](https://mobiledgex.atlassian.net/browse/EDGECLOUD-4382) as a wrapper for printing messages 
 og option Added to MobiledgeX Settings
Adding and Updating Licenses
Suppressing some warnings
Removing GetDeviceInfo from TestDeviceInfo Class


lgarner:
Wouldn't un-ticking "development" build be sufficient?

Runtime/Scripts/DeviceInfoIntegration.cs
```diff
@@ -174,9 +195,5 @@ enum NetworkDataType
   // Used for DeviceInfo in UnityEditor (any target platform)
   public class TestDeviceInfo : DeviceInfoIntegration
   {
-    public Dictionary<string, string> GetDeviceInfo()
```
franklin-huang-mobiledgex, line 177: I think this might fail to compile. I think we need a GetDeviceInfo function in order to implement the DeviceInfoIntegration interface

franklin-huang-mobiledgex (commented):
I think these logs are pretty useful... especially if a developer runs into any issues. If it's not possible to do what Garner is saying, we could write a wrapper class that allows some more configuration of the logs, so that devs can suppress them on release builds. (Or maybe use Garner's logging class in c# sdk)

Runtime/Scripts/DeviceInfoIntegration.cs
ah1053: No, it doesn't fail to complie, Actually it's giving me an warning since DeviceInfoIntegration (the Parent) have a public GetDeviceInfo method which TestDeviceInfo() inherits.

ah1053:
@lgarner  Debug.Log works in build also, but the main problem is showing all of these messaging of that in UnityEditor spams the developer console and makes it hard to find their own Debug.Log

Runtime/Scripts/MobiledgeXLogger.cs
```diff
@@ -0,0 +1,43 @@
...
+using UnityEngine;
+
+namespace MobiledgeX
+{
+    public class MobiledgeXLogger
+    {
+        public enum LogType { All, ErrorsAndWarnings, ErrorsOnly}
+        public static LogType logType;
+
+        public static void PrintWarning(string message)
```
lgarner, line 27: Although debatable, I still like "Tag" added to Loggers. It allows more creative filtering. Of course, one could also make it part of the original message, if less structured. There's obviously something like message builder if max flexibility is needed.

Also, why "Print"? These things do stream into actual logs in Unity.

Runtime/Scripts/MobiledgeXLogger.cs
```diff
@@ -0,0 +1,43 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+using UnityEngine;
+
+namespace MobiledgeX
+{
+    public class MobiledgeXLogger
```
lgarner, line 22: Maybe "Logger", since developer already chose the MobiledgeX namespace via Using to shorten the object name, but it's currently near guaranteed there's no collisions named this way.

Runtime/Scripts/DeviceInfoIntegration.cs
franklin-huang-mobiledgex: ah i think this class should implement DeviceInfo interface instead of inheriting from DeviceInfoIntegration. That way we can return dummy values from GetDeviceInfo

Runtime/Scripts/LocationService.cs
```diff
@@ -99,27 +96,39 @@ public static IEnumerator InitalizeLocationService(int maxWait = 20, bool contin
 
         /// <summary>
         /// EnsureLocation Confirm that user location is valid, user location is essential for MobiledgeX services
-        /// If Location permission is denied by User an exception will be thrown (for Android) in LocationServiceFlow()
-        ///                                                                      (for iOS) in InitalizeLocationService()
+        /// If Location permission is denied by User an exception will be thrown once RetrieveLocation() is called
         /// </summary>
         public static IEnumerator EnsureLocation()
```
franklin-huang-mobiledgex, line 101: Can we add some logs in this function so that devs know why they can't proceed with our APIs?

Runtime/Scripts/MobiledgeXSettings.cs
```diff
@@ -28,5 +28,12 @@ public class MobiledgeXSettings: ScriptableObject
         public string appVers;
         public string authPublicKey;
         public string region;
+        public MobiledgeXLogger.LogType logType = MobiledgeXLogger.LogType.ErrorsOnly;
```
franklin-huang-mobiledgex, line 31: I think default should be at least "WarningsAndErrors"

franklin-huang-mobiledgex (approved):
Cool! Logger class seems useful!

Runtime/Scripts/MobiledgeXLogger.cs
ah1053: Good point

PR #122, opened 2021-02-11
Edgecloud 4350 2.4.1

lgarner:
- Fix too vague connection error messages (RegisterClient)
- Extra guard mode to prevent Permission dialogs when not needed by application.
- DLL update needed as well (TBD).


Runtime/Scripts/MobiledgeXIntegrationHelper.cs
```diff
@@ -344,6 +344,12 @@ public async Task<bool> IsRoaming()
             string wifiIpV4 = null;
             string wifiIpV6 = null;
 
+            if (!matchingEngine.EnableEnhancedLocationServices)
```
ah1053, line 347: EnableEnhancedLocationServices will be sat in the Dll then?

lgarner: Just included the DLL with the fix to avoid the local network permissions.

ah1053 (approved):
tested on iOS 14.3 and local network  permission didn't appear.
Thank you Garner 🎩

ah1053:
Can you change the base to staging 2.5 
If you don't want to change replace Debug with Logger, I can't do it with the ChangeLog PR

PR #127, opened 2021-02-23
Connect to Multiple Apps

ah1053:
https://mobiledgex.atlassian.net/browse/EDGECLOUD-2719
Removing static variables from MobiledgeXIntegration.cs and new constructor using App Defintions.

The whole test runner suite is outdated, I will have a PR later for that ( the RunTimeTests.cs update here just to silence complier errors).

### DIFF
--- a/Runtime/Scripts/MobiledgeXIntegration.cs
+++ b/Runtime/Scripts/MobiledgeXIntegration.cs
@@ -129,17 +129,17 @@ namespace MobiledgeX
         /// <summary>
         /// Constructor for MobiledgeXIntegration. This class has functions that wrap DistributedMatchEngine functions for ease of use
         /// </summary>
-        public MobiledgeXIntegration(string OrgName , string AppName , string AppVers , string DeveloperAuthToken = "")
+        public MobiledgeXIntegration(string orgName, string appName , string appVers , string developerAuthToken = "")
         {
-            orgName = OrgName;
-            appVers = AppVers;
-            appName = AppName;
-            developerAuthToken = DeveloperAuthToken;
+            this.orgName = orgName;
+            this.appVers = appVers;
+            this.appName = appName;
+            this.developerAuthToken = developerAuthToken;
 
             // Set the platform specific way to get SIM carrier information.
             pIntegration = new PlatformIntegration();
 
-            matchingEngine = new MatchingEngine(pIntegration.CarrierInfo,pIntegration.NetInterface,pIntegration.UniqueID,pIntegration.DeviceInfo);
+            matchingEngine = new MatchingEngine(pIntegration.CarrierInfo, pIntegration.NetInterface, pIntegration.UniqueID, pIntegration.DeviceInfo);
 
             melMessaging = new MelMessaging(appName);
             matchingEngine.SetMelMessaging(melMessaging);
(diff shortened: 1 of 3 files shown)


Runtime/Scripts/MobiledgeXIntegration.cs
```diff
@@ -125,6 +125,25 @@ public MobiledgeXIntegration(CarrierInfo carrierInfo = null, NetInterface netInt
             melMessaging = new MelMessaging(appName);
             matchingEngine.SetMelMessaging(melMessaging);
         }
+        
+        /// <summary>
+        /// Constructor for MobiledgeXIntegration. This class has functions that wrap DistributedMatchEngine functions for ease of use
+        /// </summary>
+        public MobiledgeXIntegration(string OrgName , string AppName , string AppVers , string DeveloperAuthToken = "")
```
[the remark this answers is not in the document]
franklin-huang-mobiledgex: Also, I think we have been using lowercase first letter for parameters (ie. string orgName, string appName, etc.). When setting these parameters, because they are the same as your class variables, you can just do: 

```
this.orgName = orgName;
this.appVers = appVers;
this.appName = appName;
```

Runtime/Scripts/MobiledgeXIntegration.cs
```diff
@@ -125,6 +125,25 @@ public MobiledgeXIntegration(CarrierInfo carrierInfo = null, NetInterface netInt
...
+        {
+            orgName = OrgName;
+            appVers = AppVers;
+            appName = AppName;
+            developerAuthToken = DeveloperAuthToken;
+
+            // Set the platform specific way to get SIM carrier information.
+            pIntegration = new PlatformIntegration();
+
+            matchingEngine = new MatchingEngine(pIntegration.CarrierInfo,pIntegration.NetInterface,pIntegration.UniqueID,pIntegration.DeviceInfo);
```
franklin-huang-mobiledgex, line 142: nit: spaces after the commas

Tests/Runtime/MobiledgeX_RuntimeTests.cs
```diff
@@ -47,9 +47,9 @@ public void MobiledgeXEnvironmentSetup()
         public void Clean()
         {
             integration = null;
-            MobiledgeXIntegration.orgName = "";
-            MobiledgeXIntegration.appName = "";
-            MobiledgeXIntegration.appVers = "";
+            integration.orgName = "";
+            integration.appName = "";
+            integration.appVers = "";
```
franklin-huang-mobiledgex, line 52: I think you will get a null reference exception here

Tests/Runtime/MobiledgeX_RuntimeTests.cs
```diff
@@ -60,9 +60,9 @@ public void Clean()
         [TestCase("MobiledgeX", "MobiledgeX SDK Demo", "2.0")]
         public void RegisterClient(string orgName, string appName, string appVers)
         {
-            MobiledgeXIntegration.orgName = orgName;
-            MobiledgeXIntegration.appName = appName;
-            MobiledgeXIntegration.appVers = appVers;
+            integration.orgName = orgName;
+            integration.appName = appName;
+            integration.appVers = appVers;
```
franklin-huang-mobiledgex, line 65: Because you switched away from static variables, I think you can just create a MobiledgeXIntegration instance in each of your test cases instead of changing the vars each time.

franklin-huang-mobiledgex (commented):
I like the change to non-static vars! See below for some comments

Tests/Runtime/MobiledgeX_RuntimeTests.cs
ah1053: Will redo in a separate PR.

brucearmstrong (approved):

Tests/Runtime/MobiledgeX_RuntimeTests.cs
lgarner: ++

For test cases, separate independent instances just makes the tests more self contained.

PR #128, opened 2021-02-24
Run time tests

ah1053:
Updating RuntimeTests script
EDGECLOUD-4475 
https://mobiledgex.atlassian.net/wiki/spaces/SWDEV/pages/1615101953/SDK+Test+App
Screen Shot 2021-02-24 at 12 43 23 PM

### DIFF
--- a/Tests/Runtime/MobiledgeX_RuntimeTests.cs
+++ b/Tests/Runtime/MobiledgeX_RuntimeTests.cs
@@ -42,7 +42,7 @@ namespace MobiledgeX
                 Assert.Fail("MobiledgeX Plugins are not loaded in the project, Can't preform tests");
             }
         }
-        
+
         #endregion
 
         #region Run Time Tests
@@ -51,16 +51,14 @@ namespace MobiledgeX
         [TestCase("MobiledgeX-Samples", "sdktest", "9.0")]
         public void RegisterClient(string orgName, string appName, string appVers)
         {
-            MobiledgeXIntegration mix = new MobiledgeXIntegration(new CarrierInfoClass(),null,new UniqueIDClass(), new TestDeviceInfo());
-            mix.appName = appName;
-            mix.appVers = appVers;
-            mix.orgName = orgName;
+            MobiledgeXIntegration mxi = new MobiledgeXIntegration(new CarrierInfoClass(), null, new UniqueIDClass(), new TestDeviceInfo());
+            mxi.appName = appName;
+            mxi.appVers = appVers;
+            mxi.orgName = orgName;
             var task = Task.Run(async () =>
             {
-                return await RegisterHelper(mix);
+                return await RegisterHelper(mxi);
             });
-
-            Debug.Log("result of registerClient is " + task.Result);
             Assert.True(task.Result);
         }
 
@@ -68,18 +66,16 @@ namespace MobiledgeX
         [TestCase("MobiledgeX-Samples", "sdktest", "9.0")]
         public void FindCloudlet(string orgName, string appName, string appVers)
         {
-            MobiledgeXIntegration mix = new MobiledgeXIntegration(new CarrierInfoClass(), null, new UniqueIDClass(), new TestDeviceInfo());
-            mix.appName = appName;
-            mix.appVers = appVers;
-            mix.orgName = orgName;
+            MobiledgeXIntegration mxi = new MobiledgeXIntegration(new CarrierInfoClass(), null, new UniqueIDClass(), new TestDeviceInfo());
+            mxi.appName = appName;
+            mxi.appVers = appVers;
+            mxi.orgName = orgName;
             var task = Task.Run(async () =>
             {
-                bool registered = await RegisterHelper(mix);
+                bool registered = await RegisterHelper(mxi);
                 Assert.True(registered, "Unable to register");
-                return await FindCloudletHelper(mix);
+                return await FindCloudletHelper(mxi);
             });
-
-            Debug.Log("result of findCloudlet is " + task.Result);
             Assert.True(task.Result);
         }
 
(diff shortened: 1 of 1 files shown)


Tests/Runtime/MobiledgeX_RuntimeTests.cs
```diff
@@ -146,36 +156,26 @@ public void GetRestURIException(string orgName, string appName, string appVers)
...
-            });
-            // websocket reply messsage printed in console (Register event and Notification event are being emitted on the server)
-            Debug.Log(task.Result);
-            Assert.IsNotEmpty(task.Result.ToString());
-        }
-
-        [Test]
-        [TestCase("WrongCredentials", "WrongAppName", "WrongAppVersion")]
-        public void WebSocketTestExpectedException(string orgName, string appName, string appVers)
+        [TestCase("MobiledgeX-Samples", "sdktest", "9.0", -44949, 29393)]
```
brucearmstrong, line 159: Invalid lat/long values.

Tests/Runtime/MobiledgeX_RuntimeTests.cs
```diff
@@ -146,36 +156,26 @@ public void GetRestURIException(string orgName, string appName, string appVers)
...
+        public void FindCloudletFaliure(string orgName, string appName, string appVers , double longtiude, double latitude)
```
brucearmstrong, line 160: Not an error, but possible source of confusion: lat, long is more common that long, lat.

brucearmstrong (approved):
Looks good overall. Very useful stuff.

Tests/Runtime/MobiledgeX_RuntimeTests.cs
lgarner: Although internal, one can actually require a GPS style data structure to enforce field names. Or perhaps always use a specified parameter name to allow almost any parameter order.

Tests/Runtime/MobiledgeX_RuntimeTests.cs
```diff
@@ -29,107 +39,107 @@ public void MobiledgeXEnvironmentSetup()
...
 
         [Test]
-        [TestCase("MobiledgeX", "MobiledgeX SDK Demo", "2.0")]
+        [TestCase("MobiledgeX-Samples", "sdktest", "9.0")]
         public void RegisterClient(string orgName, string appName, string appVers)
         {
-            integration.orgName = orgName;
-            integration.appName = appName;
-            integration.appVers = appVers;
+            MobiledgeXIntegration mix = new MobiledgeXIntegration(new CarrierInfoClass(),null,new UniqueIDClass(), new TestDeviceInfo());
```
franklin-huang-mobiledgex, line 54: nit: spacing between commas (eg. "new CarrierInfoClass(), null, ...)

franklin-huang-mobiledgex: Also nit: "mxi" instead of "mix"

Tests/Runtime/MobiledgeX_RuntimeTests.cs
```diff
@@ -29,107 +39,107 @@ public void MobiledgeXEnvironmentSetup()
...
+            mix.orgName = orgName;
             var task = Task.Run(async () =>
             {
-                return await GetUrlHelper(orgName, appName, appVers, "http");
+                return await GetUrlHelper(mix, proto, port);
             });
-
-            Assert.AreEqual(task.Result, "http://mobiledgexsdkdemo-tcp.mobiledgexmobiledgexsdkdemo20.sdkdemo-app-cluster.us-los-angeles.gcp.mobiledgex.net:8008"); 
+            Debug.Log(task.Result);
+            Assert.IsNotEmpty(task.Result);
```
franklin-huang-mobiledgex, line 103: Can we check for the contents of the string? eg. if proto is "ws", check that the url has "ws://" and the port is there. Maybe you can have a switch statement for each of the proto cases. Same for the GetHost test

Tests/Runtime/MobiledgeX_RuntimeTests.cs
```diff
@@ -29,107 +39,107 @@ public void MobiledgeXEnvironmentSetup()
...
-            integration.appName = appName;
-            integration.appVers = appVers;
+            MobiledgeXIntegration mix = new MobiledgeXIntegration(new CarrierInfoClass(), null, new UniqueIDClass(), new TestDeviceInfo());
+            mix.appName = appName;
+            mix.appVers = appVers;
+            mix.orgName = orgName;
             var task = Task.Run(async () =>
             {
-                return await GetUrlHelper(orgName, appName, appVers, "ws");
+                Debug.Log("In get host" + proto);
```
franklin-huang-mobiledgex, line 120: nit: probably don't need this log

Tests/Runtime/MobiledgeX_RuntimeTests.cs
franklin-huang-mobiledgex: ^

Tests/Runtime/MobiledgeX_RuntimeTests.cs
franklin-huang-mobiledgex: Agree with Garner here. We do have the Loc object that you can use

Tests/Runtime/MobiledgeX_RuntimeTests.cs
```diff
@@ -188,73 +188,151 @@ public void WebSocketTestExpectedException(string orgName, string appName, strin
...
+        public void UDPTest(string orgName, string appName, string appVers, int timeOutMs)
+        {
+            MobiledgeXIntegration mix = new MobiledgeXIntegration(new CarrierInfoClass(), null, new UniqueIDClass(), new TestDeviceInfo());
+            mix.appName = appName;
+            mix.appVers = appVers;
+            mix.orgName = orgName;
+            var task = Task.Run(async () =>
+            {
+                string hostName= await GetHostHelper(mix, "udp");
+                Debug.Log("hosdasdsa" + hostName);
```
franklin-huang-mobiledgex, line 225: nit: extra log message

franklin-huang-mobiledgex (commented):
Awesome! See below for mostly nits

Tests/Runtime/MobiledgeX_RuntimeTests.cs
ah1053: Fixed I checked for proto and port in the url created

Tests/Runtime/MobiledgeX_RuntimeTests.cs
ah1053: I don't really need Loc object since I am using setFallbackLocation.
When it comes to lat long instead of long lat, I agree but this might need to happen in another PR in MobiledgeXIntegration setFallBackLocation().

LeonAdams (approved):
Checked the app instances currently up and found them working as expected.
Thanks! This will help me out quite a bit!

LeonAdams (approved):

PR #140, opened 2021-03-09
Readme Removing DMEDNS Exception

ah1053:


lgarner (approved):
For completeness, InnerException needs to be examined.

PR #151, opened 2021-05-25
Remove SDK Fix

ah1053:
EDGECLOUD-4789


LeonAdams (approved):
I like the restart ..... just need to see that it works ...    ;-)

PR #156, opened 2021-06-18
EdgeEvents Updates

ah1053:
- Adding PerformanceMarginSwitch, AutoMigration
- Some Indentation for EdgeEventsConfig
- Updating Handling Events Implementation

### DIFF
--- a/Runtime/Scripts/EdgeEventsConfig.cs
+++ b/Runtime/Scripts/EdgeEventsConfig.cs
@@ -26,9 +26,9 @@ namespace MobiledgeX
   public class EdgeEventsConfig
   {
     /// <summary>
-    /// Latency threshold (In Milliseconds) as the limit to automatically migrate to a new cloudlet
+    /// Latency threshold (In Milliseconds) is the threshold at which MobiledgeX SDK will start to look for a cloudlet with better latency
     /// </summary>
-    [Tooltip("Latency threshold (In Milliseconds) as the limit to automatically migrate to a new cloudlet")]
+    [Tooltip("Latency threshold (In Milliseconds) is the threshold at which MobiledgeX SDK will start to look for a cloudlet with better latency")]
     public double latencyThresholdTriggerMs;
     /// <summary>
     /// Port information for latency testing
@@ -48,15 +48,15 @@ namespace MobiledgeX
     [Tooltip("Config for location updates")]
     public UpdateConfig locationConfig;
     /// <summary>
-    /// List of triggers that application that will trigger a new find cloudlet.
+    /// List of triggers that will trigger a new find cloudlet.
     /// </summary>
-    [Tooltip("List of triggers that application that will trigger a new find cloudlet.")]
+    [Tooltip("List of triggers that will trigger a new find cloudlet.")]
     public List<FindCloudletEventTrigger> newFindCloudletEventTriggers;
 
     /// <summary>
-    /// Allow MobiledgeX EdgeEvents to automatically connect to the new Cloudlet received from the DME server
+    /// Allow MobiledgeX EdgeEvents to automatically stop the current EdgeEvents connection and start a new EdgeEvents connection to receive events from the new cloudlet
     /// </summary>
-    [Tooltip("Allow MobiledgeX EdgeEvents to automatically connect to the new Cloudlet received from the DME server.")]
+    [Tooltip("Allow MobiledgeX EdgeEvents to automatically stop the current EdgeEvents connection and start a new EdgeEvents connection to receive events from the new cloudlet")]
     public bool autoMigration;
     /// <summary>
     /// Average performance must be by better by this latency margin (0 to 1.0f) before notifying of switch.
@@ -88,9 +88,9 @@ namespace MobiledgeX
          "\nWorks only if the UpdatePattern is set to OnInterval")]
     public int updateIntervalSeconds;
     /// <summary>
-    /// Maximum number of updates through out the App lifetime.
+    /// Maximum number of updates throughout the App lifetime.
     /// <para>Works only if the UpdatePattern is set to <b>OnInterval</b></para>
-    /// <para>Set to -1 for updates to run till the EdgeEvents connection is closed</para>
+    /// <para>Set to 0 for updates to run till the EdgeEvents connection is closed</para>
     /// </summary>
     [Tooltip("Maximum number of updates through out the App lifetime." +
         "\nWorks only if the UpdatePattern is set to OnInterval" +
--- a/Runtime/Scripts/EdgeEventsConfig.cs
+++ b/Runtime/Scripts/EdgeEventsConfig.cs
@@ -1,123 +1,157 @@
-﻿using System;
+﻿/**
+ * Copyright 2018-2021 MobiledgeX, Inc. All rights and licenses reserved.
+ * MobiledgeX, Inc. 156 2nd Street #408, San Francisco, CA 94105
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+using System;
 using UnityEngine;
 using DistributedMatchEngine;
 using System.Collections.Generic;
 
 namespace MobiledgeX
 {
-    [Serializable]
-    public struct EdgeEventsConfig
-    {
-        /// <summary>
-        /// Latency threshold (In Milliseconds) as the limit to automatically migrate to a new cloudlet
-        /// </summary>
-        [Tooltip("Latency threshold (In Milliseconds) as the limit to automatically migrate to a new cloudlet")]
-        public double latencyThresholdTriggerMs;
-        /// <summary>
-        /// Port information for latency testing
-        /// <para>Use 0 to select a random port</para>
-        /// </summary>
-        [Tooltip("Port information for latency testing" +
-            "\nUse 0 to select a random port")]
-        public int latencyTestPort;
-        /// <summary>
-        /// Config for latency updates
-        /// </summary>
-        [Tooltip("Config for latency updates")]
-        public UpdateConfig latencyConfig;
-        /// <summary>
-        /// Config for location updates
-        /// </summary>
-        [Tooltip("Config for location updates")]
-        public UpdateConfig locationConfig;
+  [Serializable]
+  public class EdgeEventsConfig
+  {
+    /// <summary>
+    /// Latency threshold (In Milliseconds) as the limit to automatically migrate to a new cloudlet
+    /// </summary>
+    [Tooltip("Latency threshold (In Milliseconds) as the limit to automatically migrate to a new cloudlet")]
+    public double latencyThresholdTriggerMs;
+    /// <summary>
+    /// Port information for latency testing
+    /// <para>Use 0 to select a random port</para>
+    /// </summary>
+    [Tooltip("Port information for latency testing" +
+        "\nUse 0 to select a random port")]
+    public int latencyTestPort;
+    /// <summary>
+    /// Config for latency updates
+    /// </summary>
+    [Tooltip("Config for latency updates")]
+    public UpdateConfig latencyConfig;
+    /// <summary>
+    /// Config for location updates
+    /// </summary>
+    [Tooltip("Config for location updates")]
+    public UpdateConfig locationConfig;
+    /// <summary>
+    /// List of triggers that application that will trigger a new find cloudlet.
+    /// </summary>
+    [Tooltip("List of triggers that application that will trigger a new find cloudlet.")]
+    public List<FindCloudletEventTrigger> newFindCloudletEventTriggers;
 
-        /// <summary>
-        /// List of triggers that will trigger a new find cloudlet.
-        /// </summary>
-        [Tooltip("List of triggers that will trigger a new find cloudlet.")]
-        public List<FindCloudletEventTrigger> newFindCloudletEventTriggers;
-    }
+    /// <summary>
+    /// Allow MobiledgeX EdgeEvents to automatically connect to the new Cloudlet received from the DME server
+    /// </summary>
+    [Tooltip("Allow MobiledgeX EdgeEvents to automatically connect to the new Cloudlet received from the DME server.")]
+    public bool autoMigration;
+    /// <summary>
+    /// Average performance must be by better by this latency margin (0 to 1.0f) before notifying of switch.
+    /// </summary>
+    [Tooltip("Average performance must be by better by this latency margin (0 to 1.0f) before notifying of switch.")]
+    public float performanceSwitchMargin = 0.05f;
+  }
 
-    [Serializable]
-    public struct UpdateConfig
-    {
-        /// <summary>
-        /// UpdatePattern for sending client events
-        /// <para><b>OnInterval</b> update every updateInterval seconds </para>
-        /// <para><b>OnStart</b> only update once the connection starts</para>
-        /// <para><b>OnTrigger</b> the application is responsible for sending the events </para>
-        /// </summary>
-        [Tooltip(" UpdatePattern for sending client events" +
-            "\nOnInterval: update every updateInterval seconds" +
-            "\nOnStart: only update once the connection starts" +
-            "\nOnTrigger: the application is responsible for sending the events ")]
-        public UpdatePattern updatePattern;
-        /// <summary>
-        /// Update interval in seconds
-        /// <para>Works only if the UpdatePattern is set to <b>OnInterval</b></para>
-        /// </summary>
-        [Tooltip("Update interval in seconds" +
-             "\nWorks only if the UpdatePattern is set to OnInterval")]
-        public int updateIntervalSeconds;
-        /// <summary>
-        /// Maximum number of updates through out the App lifetime.
-        /// <para>Works only if the UpdatePattern is set to <b>OnInterval</b></para>
-        /// <para>Set to -1 for updates to run till the EdgeEvents connection is closed</para>
-        /// </summary>
-        [Tooltip("Maximum number of updates through out the App lifetime." +
-            "\nWorks only if the UpdatePattern is set to OnInterval" +
-            "\nSet to 0 for updates to run till the EdgeEvents connection is closed")]
-        public int maxNumberOfUpdates;
-    }
 
-    //TODO add comments
-    public struct FindCloudletEvent
-    {
-        public FindCloudletReply newCloudlet;
-        public FindCloudletEventTrigger trigger;
-    }
-
+  [Serializable]
+  public class UpdateConfig
+  {
     /// <summary>
     /// UpdatePattern for sending client events
+    /// <para><b>OnInterval</b> update every updateInterval seconds </para>
+    /// <para><b>OnStart</b> only update once the connection starts</para>
+    /// <para><b>OnTrigger</b> the application is responsible for sending the events </para>
     /// </summary>
-    public enum UpdatePattern
-    {
-        OnInterval=0,
-        OnStart=1,
-        OnTrigger=2,
-    }
+    [Tooltip(" UpdatePattern for sending client events" +
+        "\nOnInterval: update every updateInterval seconds" +
+        "\nOnStart: only update once the connection starts" +
+        "\nOnTrigger: the application is responsible for sending the events ")]
+    public UpdatePattern updatePattern;
     /// <summary>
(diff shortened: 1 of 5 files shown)


Runtime/Scripts/PersistentConnection.cs
```diff
@@ -131,192 +218,176 @@ IEnumerator OnIntervalEdgeEventsLocation(EdgeEventsConnection connection)
...
-      StopAllCoroutines();
+      latencyUpdatesCounter = 0;
+      latencyUpdatesRunning = false;
+      locationUpdatesCounter = 0;
+      locationUpdatesRunning = false;
+      location = null;
     }
 
-    private void OnDestroy()
+    void PropagateError(FindCloudletEventTrigger trigger, string error_msg)
```
lgarner, line 389: Is there a better name? Every error a FindCloudletEventTrigger.

Runtime/Scripts/PersistentConnection.cs
```diff
@@ -19,50 +38,122 @@ public class PersistentConnection : MonoBehaviour
...
+    }
+
+    private void OnApplicationQuit()
+    {
+      StopAllCoroutines();// stop edge events streaming
+    }
+
+    private void OnDestroy()
+    {
+      StopAllCoroutines();// stop edge events streaming
```
lgarner, line 90: Ok. This can be called more than once?

Runtime/Scripts/MobiledgeXSettings.cs
```diff
@@ -34,6 +34,7 @@ public class MobiledgeXSettings: ScriptableObject
 
         private void OnEnable()
         {
+            edgeEventsConfig.newFindCloudletEventTriggers.Add(FindCloudletEventTrigger.Error);
```
lgarner, line 37: HashSet work here?

Runtime/Scripts/EdgeEventsConfig.cs
```diff
@@ -1,123 +1,157 @@
...
+    public UpdateConfig latencyConfig;
+    /// <summary>
+    /// Config for location updates
+    /// </summary>
+    [Tooltip("Config for location updates")]
+    public UpdateConfig locationConfig;
+    /// <summary>
+    /// List of triggers that application that will trigger a new find cloudlet.
+    /// </summary>
+    [Tooltip("List of triggers that application that will trigger a new find cloudlet.")]
```
brucearmstrong, line 53: Kind of awkward. Suggestion:
List of triggers for the application to perform a new find cloudlet.

Runtime/Scripts/PersistentConnection.cs
```diff
@@ -131,192 +217,193 @@ IEnumerator OnIntervalEdgeEventsLocation(EdgeEventsConnection connection)
...
+      double normalizedLatency = receivedStats.Avg - (receivedStats.Avg * config.performanceSwitchMargin);
+      if (site.average < normalizedLatency)
       {
-        StartCoroutine(OnIntervalEdgeEventsLatency(integration.matchingEngine.EdgeEventsConnection, integration.GetHost()));
+        PropagateSuccess(FindCloudletEventTrigger.LatencyTooHigh, integration.latestFindCloudletReply);
+      }
+      else
+      {
+        PropagateError(FindCloudletEventTrigger.LatencyTooHigh,
+          "Cloudlet obtained from FindCloudletPerformanceMode have higher latency than the previous cloudlet");
```
brucearmstrong, line 352: nit: "has higher"

lgarner (approved):
Ok, with comments.

Runtime/Scripts/PersistentConnection.cs
```diff
@@ -19,50 +38,121 @@ public class PersistentConnection : MonoBehaviour
...
+      integration.matchingEngine.EdgeEventsReceiver += HandleReceivedEvents;
       EdgeEventsConnection connection = integration.matchingEngine.EdgeEventsConnection;
-      
+      bool valid = ValidateConfigs();
+      if (!valid)
+      {
+        return;
+      }
       location = LocationService.RetrieveLocation();
       AppPort appPort = integration.GetAppPort(LProto.Tcp, config.latencyTestPort == 0 ? 0 : config.latencyTestPort);
```
franklin-huang-mobiledgex, line 157: the ternary operator is unnecessary. can just be:

`AppPort appPort = integration.GetAppPort(LProto.Tcp, config.latencyTestPort);`

Runtime/Scripts/PersistentConnection.cs
```diff
@@ -131,192 +217,193 @@ IEnumerator OnIntervalEdgeEventsLocation(EdgeEventsConnection connection)
...
+      yield return StartCoroutine(LocationService.EnsureLocation());
+      if (hasTCPPorts)
+      {
+        connection.TestConnectAndPostLatencyUpdate(host, (uint)config.latencyTestPort, location).ConfigureAwait(false);
+      }
+      else
+      {
+        integration.matchingEngine.EdgeEventsConnection
+          .TestPingAndPostLatencyUpdate
+          (integration.GetHost(), location).ConfigureAwait(false);
```
franklin-huang-mobiledgex, line 256: Can this just be `connection.TestPingAndPostLatencyUpdate` like it is on line 250?

Runtime/Scripts/MobiledgeXSettings.cs
franklin-huang-mobiledgex: ^

Runtime/Scripts/PersistentConnection.cs
```diff
@@ -131,192 +217,193 @@ IEnumerator OnIntervalEdgeEventsLocation(EdgeEventsConnection connection)
...
-            connection.TestConnectAndPostLatencyUpdate(host, (uint)config.latencyTestPort, location).ConfigureAwait(false);
+            await integration.matchingEngine.EdgeEventsConnection.TestConnectAndPostLatencyUpdate(host, (uint)config.latencyTestPort, location);
           }
           else
           {
-            integration.matchingEngine.EdgeEventsConnection
-              .TestPingAndPostLatencyUpdate
-              (integration.GetHost(), location).ConfigureAwait(false);
+            await integration.matchingEngine.EdgeEventsConnection.TestPingAndPostLatencyUpdate(host, location);
           }
```
franklin-huang-mobiledgex, line 297: I don't think we have to do anything if config.newFindCloudletEventTriggers does not contain the trigger. EventLatencyRequest is most likely sent by the developer and they are also setting the FIndCloudletTriggers. So, if "EventLatencyRequest" is not set here, then we should assume that they don't want latency tests at all

Runtime/Scripts/PersistentConnection.cs
```diff
@@ -131,192 +217,193 @@ IEnumerator OnIntervalEdgeEventsLocation(EdgeEventsConnection connection)
...
-          latencyUpdatesCounter++;
-          yield return StartCoroutine(OnIntervalEdgeEventsLatency(connection, host));
         }
+      }
+    }
 
-    async void HandleServerReceivedEvents(ServerEdgeEvent edgeEvent)
+    async void ProcessLatency(Statistics stats)
+    {
+      FindCloudletReply currentCloudlet = integration.latestFindCloudletReply;
```
franklin-huang-mobiledgex, line 304: Ah, so using the latestFindCloudletReply was an issue that Garner and I resolved in https://mobiledgex.atlassian.net/browse/EDGECLOUD-5100. Since, you are setting latestFindCloudletReply after every FindCloudlet call, there is a case where the FindCloudletPerformance call returns a cloudlet that we will not use because it is not better than the performanceMargin. Then the latestFindCloudletReply will be set to a cloudlet that the client is not actually on. So, if the client actually moves closer to that cloudlet in the future, it will compare it to latestFindCloudletReply (which will be the same) and not switch to it when it is supposed to.

The fix we made was to keep track of currentFindCloudletReply as an EdgeEventsConnection class variable (or PersistentConnection in your case) and that will be the current cloudlet that we are receiving events about.

Runtime/Scripts/PersistentConnection.cs
```diff
@@ -131,192 +217,193 @@ IEnumerator OnIntervalEdgeEventsLocation(EdgeEventsConnection connection)
...
+      if (stats.Avg < config.latencyThresholdTriggerMs)
+      {
+        integration.UseFindCloudletPerformanceMode(true);
```
franklin-huang-mobiledgex, line 307: hmm. We should reset the FindCloudletPerformanceMode after the FIndCloudlet finishes:

```
prevPerformanceMode = integration.PerformanceMode;
integration.UseFindCloudletPerformanceMode(true);
bool fcResult = await integration.FindCloudlet();
intergration.UseFindCloudletPerformanceMode(prevPerformanceMode);
```

Runtime/Scripts/PersistentConnection.cs
```diff
@@ -131,192 +217,193 @@ IEnumerator OnIntervalEdgeEventsLocation(EdgeEventsConnection connection)
...
-    private void OnApplicationFocus(bool focus)
+    async void CompareLatencies(Statistics receivedStats)
     {
-      if (locationUpdatesRunning)
+      AppPort appPort;
+      Site site;
+      var netTest = new NetTest(integration.matchingEngine);
+      if (hasTCPPorts)
+      {
+        appPort = integration.GetAppPort(LProto.Tcp, config.latencyTestPort == 0 ? 0 : config.latencyTestPort);
```
franklin-huang-mobiledgex, line 335: same comment as above about ternary operator

Runtime/Scripts/PersistentConnection.cs
franklin-huang-mobiledgex: nit: this comment should be more like: "Latency threshold exceeded, but no other cloudlets have a meaningful improvement in latency"

Runtime/Scripts/PersistentConnection.cs
```diff
@@ -131,192 +217,193 @@ IEnumerator OnIntervalEdgeEventsLocation(EdgeEventsConnection connection)
...
+      findCloudletEvent.newCloudlet = newCloudlet;
+      EdgeEventsStatus eventStatus = new EdgeEventsStatus(Status.success);
+      try
+      {
+        integration.NewFindCloudletHandler(eventStatus, findCloudletEvent);
+      }
+      catch (NullReferenceException)
+      {
+        Debug.LogError("Missing NewCloudletHandler, " +
+          "Listen to EdgeEvents by using \"mobiledgexIntegration.NewCloudletHandler += HandleNewFindCloudlet;\"");
```
franklin-huang-mobiledgex, line 379: You should probably validate that NewCloudletHandler exists in your validate function as well

franklin-huang-mobiledgex (commented):
Looking pretty good!

brucearmstrong (approved):

Runtime/Scripts/PersistentConnection.cs
ah1053: Maybe I can update the name since you don't like it (I am curious why? Is not representative of the method functionality!) 
Would something PropagateEdgeEventError be more coherent with the context?

Runtime/Scripts/PersistentConnection.cs
ah1053: Probably not, Since the scene or the game will be ended .
Also a good reminder that this object should be active with scenes transitions at least Additive Scene Loading.

Runtime/Scripts/PersistentConnection.cs
ah1053: I researched this and probably I can't validate that before triggering an event. 
@lgarner  Do you know of a way that I can check of the Action Listeners Count or Existance **without bothering the developer with an event**-something that is not a performance killer **no reflection** -?

Runtime/Scripts/PersistentConnection.cs
lgarner: I'm actually curious if we really want it a boolean, instead of the Mode, which is a public state Enum.

Runtime/Scripts/PersistentConnection.cs
lgarner: I am referring to the same StopAllCroutines() is called twice in the "Decommissioning" phase of the app: OnApplicationQuit() --> OnDisable() --> OnDestroy().

Runtime/Scripts/PersistentConnection.cs
```diff
@@ -19,50 +38,121 @@ public class PersistentConnection : MonoBehaviour
...
+
+    private void OnDestroy()
+    {
+      StopAllCoroutines();// stop edge events streaming
+    }
+    #endregion
+
+    #region PersistentConnection Functions
+
+    bool ValidateConfigs()
```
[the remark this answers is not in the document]
lgarner: Is possible the developer doesn't want even those for some reason during the app lifecycle (even though they can ignore it contextually without needing us to enforce it).

Runtime/Scripts/PersistentConnection.cs
```diff
@@ -19,50 +38,122 @@ public class PersistentConnection : MonoBehaviour
...
+    {
+      if (config.newFindCloudletEventTriggers.Count == 1)//No FindCloudletTrigger except Error (Added By Default)
```
lgarner, line 98: Nit: Space, or rather, triggers.contains(Error)"

Runtime/Scripts/PersistentConnection.cs
lgarner: Given there's one for now, perhaps we can just rely on method name overloading for now to distinguish. Fine as is then.

Runtime/Scripts/PersistentConnection.cs
ah1053: Aha, Ok.
Yeah StopAllCoroutines() can be called more than once, like the following:

OnEnable() -> StartEdgeEventsCoroutines() -> OnApplicationPause(true) -> **StopAllCoroutines()** ->  OnApplicationPause(false) -> StartEdgeEventsCoroutines() -> OnDestroy() -> **StopAllCoroutines()** -> ||

Runtime/Scripts/PersistentConnection.cs
ah1053: If you are talking about UseFindCloudletPerformanceMode(true)

For setting FindCloudletMode we have a Property with a getter (Mode) and a private variable (mode) where we change the mode internally.

From previous conversations, I think we agreed to expose UseFindCloudletPerformanceMode(true) as the developer way for using PerformanceMode aligning with UseWifiOnly(boolean).

I am just going to change the variable (mode) encapsulation from private to internal so I can access its value it from PresistentConnection.cs

Screen Shot 2021-06-21 at 12 56 39 PM

Runtime/Scripts/PersistentConnection.cs
ah1053: Fixed to Ignore LatencyRequest if LatencyTooHigh is not a trigger

Runtime/Scripts/PersistentConnection.cs
ah1053: Just an explanation, EventTrigger.Error is added by default, I am checking if there is any other triggers added.

Runtime/Scripts/PersistentConnection.cs
ah1053: Will do this in another PR if I found a solution.

Runtime/Scripts/MobiledgeXIntegration.cs
```diff
@@ -75,15 +75,15 @@ public partial class MobiledgeXIntegration : IDisposable
...
+        internal FindCloudletReply latestFindCloudletReply = null; // Stored to be used in GetUrl, GetHost, GetPort, Get[]Connection
         bool latestVerifyLocationStatus = false; // Whether the most recent verifyLocation call was successful
         FindCloudletMode mode = FindCloudletMode.PROXIMITY; // FindCloudlet mode
         AppPort latestAppPort = null;
         AppPort[] latestAppPortList = null;
         Location fallbackLocation = new Location(0,0);
         CarrierInfoClass carrierInfoClass = new CarrierInfoClass(); // used for IsRoaming check
         MelMessaging melMessaging;
-        PersistentConnection persistentConnection;
+        internal PersistentConnection persistentConnection;
```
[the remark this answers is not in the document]
ah1053: Will do in a future PR once I find an appropriate name.

Runtime/Scripts/MobiledgeXSettings.cs
ah1053: Will consider in the next PR

Runtime/Scripts/PersistentConnection.cs
```diff
@@ -19,52 +39,155 @@ public class PersistentConnection : MonoBehaviour
...
+      }
+
+      if (config.performanceSwitchMargin > 1 || config.performanceSwitchMargin < 0)
+      {
+        PropagateError(FindCloudletEventTrigger.Error, "performanceSwitchMargin must between (0 to 1.0f)");
+        return false;
+      }
+
+      if (config.newFindCloudletEventTriggers.Contains(FindCloudletEventTrigger.LatencyTooHigh) ||
+        config.newFindCloudletEventTriggers.Contains(FindCloudletEventTrigger.CloserCloudlet))
```
franklin-huang-mobiledgex, line 118: I wouldn't use an `||` here. These should be checked individually:

```
if (config.newFindCloudletEventTriggers.Contains(FindCloudletEventTrigger.LatencyTooHigh) 
{
// validate latencyConfig
}

if (config.newFindCloudletEventTriggers.Contains(FindCloudletEventTrigger.CloserCloudlet)
{
// validate locationConfig
}

```

Runtime/Scripts/PersistentConnection.cs
franklin-huang-mobiledgex: hmm i'm confused. it looks like the code from line 321 to line 333 is still there. that can all be removed

Runtime/Scripts/PersistentConnection.cs
franklin-huang-mobiledgex: has this been fixed?

Runtime/Scripts/PersistentConnection.cs
franklin-huang-mobiledgex: has this been fixed?

Runtime/Scripts/PersistentConnection.cs
franklin-huang-mobiledgex: this should be an easy fix

Runtime/Scripts/PersistentConnection.cs
franklin-huang-mobiledgex: please file a jira ticket to track this

Runtime/Scripts/MobiledgeXSettings.cs
franklin-huang-mobiledgex: please file a jira ticket for this

Runtime/Scripts/MobiledgeXIntegration.cs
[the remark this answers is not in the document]
franklin-huang-mobiledgex: please file a jira ticket for this

franklin-huang-mobiledgex (commented):
please make fixes or link to jira tickets before resolving conversations

Runtime/Scripts/MobiledgeXIntegration.cs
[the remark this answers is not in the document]
ah1053: will do.

Runtime/Scripts/PersistentConnection.cs
ah1053: Yes I am passing the currentCloudlet to CompareLatencies to revert to old cloudlet if the new cloudlet is not better.
Let me know if this is suffcient.

Runtime/Scripts/PersistentConnection.cs
ah1053: Yes, it have been resolved as mentioned in my comment. Let me know if this is sufficient.

Runtime/Scripts/PersistentConnection.cs
ah1053: Done https://mobiledgex.atlassian.net/browse/EDGECLOUD-5172

Runtime/Scripts/PersistentConnection.cs
ah1053: Will need to fix that.

Runtime/Scripts/PersistentConnection.cs
ah1053: Will need to fix that.

Runtime/Scripts/PersistentConnection.cs
franklin-huang-mobiledgex: i think you should set `integration.mode = prevMode;` right after the FindCloudlet call (ie. line 362)

ah1053:
Added Tickets for:
1. Renaming PersistentConneciton https://mobiledgex.atlassian.net/browse/EDGECLOUD-5182
2. Validate EdgeEventsConnection have a NewCloudletHandler https://mobiledgex.atlassian.net/browse/EDGECLOUD-5172
3. HashSet instead of List for newFindCloudletTriggers (Backlog) https://mobiledgex.atlassian.net/browse/EDGECLOUD-5183
Working on a PR now for other fixes.

PR #159, opened 2021-06-25
Fix Zero Latency Port

ah1053:
1. persistentConnection public for overloading EdgeEventsLocation.
2. Capture Location at Start Coroutine.
3. Better log message for the current config.
4. Use any available TCP port if config.latencyPort is zero.
5. Modified Impl. for Opening the connection to avoid crashes
6. StopEdgeEvents() for Terminating EdgeEventsConnection.


Runtime/Scripts/PersistentConnection.cs
```diff
@@ -408,14 +506,16 @@ async void CompareLatencies(Statistics receivedStats)
         PropagateError(FindCloudletEventTrigger.LatencyTooHigh,
           "Latency threshold exceeded, but no other cloudlets have a meaningful improvement in latency");
       }
+      return;
     }
 
-    async Task StopEdgeEvents()
+    void StopEdgeEvents()
```
lgarner, line 512: This intended? This makes StopEdgeEvents synchronous with Close() (UI Pause if not in background thread).

Runtime/Scripts/PersistentConnection.cs
```diff
@@ -296,104 +327,176 @@ IEnumerator OnIntervalEdgeEventsLatency(EdgeEventsConnection connection, string
...
+          }
+          return;
+        case ServerEventType.EventLatencyProcessed:
+          if (fcTriggers.Contains(FindCloudletEventTrigger.LatencyTooHigh))
+          {
+            latestServerStats = edgeEvent.Statistics;
+            if(!latencyProcessingInProgress)
             {
-              await RespondToLatencyRequest();
+              startLatencyProcessing = true;
```
franklin-huang-mobiledgex, line 356: why do we need this boolean? Is it not possible to call ProcessLatency from this function instead of from Update?

Runtime/Scripts/PersistentConnection.cs
```diff
@@ -296,104 +327,176 @@ IEnumerator OnIntervalEdgeEventsLatency(EdgeEventsConnection connection, string
...
+          {
+            if (edgeEvent.NewCloudlet == null)
             {
-              PropagateSuccess(trigger, edgeEvent.NewCloudlet);
+              PropagateError(FindCloudletEventTrigger.CloudletMaintenanceStateChanged, "Received CloudletStateChanged Event but NewCloudlet is null");
+              return;
             }
-            return;
-        }
+            PropagateSuccess(FindCloudletEventTrigger.CloudletStateChanged, edgeEvent.NewCloudlet);
```
franklin-huang-mobiledgex, line 379: this should be  `PropagateSuccess(FindCloudletEventTrigger.CloudletMaintenanceStateChanged, edgeEvent.NewCloudlet);`

Runtime/Scripts/PersistentConnection.cs
ah1053: Yes this is intended.
On UI Pause we StopAllCoroutines()

Screen Shot 2021-06-29 at 11 09 05 AM
 

StopEdgeEvents () is for Closing the connection (from our convo on slack about one function for reseting variables and terminating the EdgeEvents Connection)

ah1053: Does this answer your question?

Runtime/Scripts/PersistentConnection.cs
```diff
@@ -296,104 +327,176 @@ IEnumerator OnIntervalEdgeEventsLatency(EdgeEventsConnection connection, string
...
             {
-              await RespondToLatencyRequest();
+              startLatencyProcessing = true;
+            }
+          }
+          return;
+        case ServerEventType.EventAppinstHealth:
+          if (fcTriggers.Contains(FindCloudletEventTrigger.AppInstHealthChanged))
+          {
+            if (edgeEvent.NewCloudlet == null)
```
franklin-huang-mobiledgex, line 363: sorry i wasn't clear. but all these checks should also have a check to see if edgeEvent.ErrorMsg != "" (ie. `if (edgeEvent.NewCloudlet == null || edgeEvent.ErrorMsg != "")`). Then you can add that errormsg to your ProgagateError message

Runtime/Scripts/PersistentConnection.cs
```diff
@@ -296,104 +327,176 @@ IEnumerator OnIntervalEdgeEventsLatency(EdgeEventsConnection connection, string
...
+          }
+          return;
+        case ServerEventType.EventError:
+          Logger.Log("Received EventError from server, " + edgeEvent.ToString());
+          PropagateError(FindCloudletEventTrigger.Error, edgeEvent.ErrorMsg);
+          return;
+        default:
+        case ServerEventType.EventUnknown:
+          Logger.Log("Received Unknown event, "+edgeEvent.ToString());
+          return;
```
franklin-huang-mobiledgex, line 411: these can be consolidated into the default case

Runtime/Scripts/PersistentConnection.cs
```diff
@@ -38,24 +38,23 @@ public class PersistentConnection : MonoBehaviour
...
     int latencyUpdatesCounter;
     int locationUpdatesCounter;
-    Dictionary<ServerEventType, FindCloudletEventTrigger> event_trigger_dict;
     FindCloudletReply currentFindCloudlet;
     public Loc location;
     public bool useMobiledgexLocationServices = true;
+    static bool stopEdgeEvents;
+    static bool startLatencyProcessing;
+    static bool latencyProcessingInProgress;
+    static Statistics latestServerStats;
```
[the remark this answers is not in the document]
franklin-huang-mobiledgex: Also, i'm not sure i see the point in keeping latestServerStats. If you are using it to compare FindCloudletPerformance results, I think that can be kept to the scope of the function (ProcessLatency, probably)

Runtime/Scripts/PersistentConnection.cs
```diff
@@ -296,104 +327,176 @@ IEnumerator OnIntervalEdgeEventsLatency(EdgeEventsConnection connection, string
...
-        integration.UseFindCloudletPerformanceMode(true);
-        bool fcResult = await integration.FindCloudlet();
-        if (fcResult)
+      Logger.Log("Processing Latency Received from Server");
+      if (stats.Avg > config.latencyThresholdTriggerMs)
+      {
+        Logger.Log("Current Latency (" + stats.Avg + ") > LatencyThreshold (" + config.latencyThresholdTriggerMs + ")");
+        Logger.Log("Performing FindCloudlet PerformanceMode");
+        MobiledgeXIntegration tempMxi = new MobiledgeXIntegration();
+        bool regResult = await tempMxi.Register("eu-qa.dme.mobiledgex.net", 50051);
```
franklin-huang-mobiledgex, line 439: hmm. the host shouldn't be hardcoded. It should be the same dme that the persistent connection uses (same for your FindCloudlet call)

Runtime/Scripts/PersistentConnection.cs
```diff
@@ -296,104 +327,176 @@ IEnumerator OnIntervalEdgeEventsLatency(EdgeEventsConnection connection, string
...
+        if (regResult)
```
franklin-huang-mobiledgex, line 440: nit: it's a little bit confusing to have a bunch of these nested if/elses. It might be easier to understand if you check for failures and return before moving on. ie.

```
if (!regResult)
{
    // propagate error and return
}
... find cloudlet code ...
if (!fcSuccess)
{
    // propagate error and return
}
... rest of code...
```

franklin-huang-mobiledgex (commented):
good progress! see below for comments

Runtime/Scripts/PersistentConnection.cs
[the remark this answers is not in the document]
lgarner: Yes, maybe internal or private, as they aren't even visible, unless for something like process wide global state.

Runtime/Scripts/PersistentConnection.cs
```diff
@@ -93,6 +92,24 @@ private void OnDestroy()
     {
       StopAllCoroutines();// stop edge events streaming
     }
+
+    private void Update()
+    {
+      if (stopEdgeEvents)
+      {
+        stopEdgeEvents = false;
```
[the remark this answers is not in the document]
lgarner: "stopEdgeEventsRequested"?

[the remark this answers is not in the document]
lgarner: Or { open, closing, closed } or some more structured enum variant as below.

Runtime/Scripts/PersistentConnection.cs
```diff
@@ -241,14 +260,19 @@ void StartEdgeEventsLocation(EdgeEventsConnection connection)
 
     void StartEdgeEventsLatency(EdgeEventsConnection connection, string host)
     {
+      Logger.Log("EdgeEvents Posting latency update, Location to send [" + location.Latitude + ", " + location.Longitude + "]");
+      bool requestSent;
       if (hasTCPPorts)
       {
-        connection.TestConnectAndPostLatencyUpdate(host, (uint)config.latencyTestPort, location).ConfigureAwait(false);
+        requestSent = connection.TestConnectAndPostLatencyUpdate(host, (uint)config.latencyTestPort, location).ConfigureAwait(false).GetAwaiter().GetResult();
```
lgarner, line 267: Or skip await since it's mostly intended to be fire and forget? (or use await <task>.ConfigureAwait(false) to actually wait on current thread for the background task).

Runtime/Scripts/PersistentConnection.cs
```diff
@@ -275,7 +299,7 @@ IEnumerator OnIntervalEdgeEventsLocation(EdgeEventsConnection connection)
       }
       yield return new WaitForSecondsRealtime(config.locationConfig.updateIntervalSeconds);
       yield return StartCoroutine(UpdateLocation());
-      Logger.Log("EdgeEvents Posting location update, Location to send [" + location.Latitude + ", " + location.Longitude + "]");
+      Logger.Log("EdgeEvents Posting location update, Location to send ["+location.Latitude+", "+ location.Longitude+"]");
```
lgarner, line 302: Nit: spaces looked correct prior.

Runtime/Scripts/PersistentConnection.cs
lgarner: (Note: EdgeEventsConnection actually stored DME host until Close() wipes it, may need to update).

Runtime/Scripts/PersistentConnection.cs
lgarner: Ok, the second context for calling it is App directed cloudlet migration.

Runtime/Scripts/PersistentConnection.cs
```diff
@@ -442,24 +542,23 @@ async void PropagateSuccess(FindCloudletEventTrigger trigger, FindCloudletReply
...
       {
-        await StopEdgeEvents();
+        StopEdgeEvents();
         integration.persistentConnection.startStreamingEvents(integration);
       }
       else
       {
         //Migration occurs and autoMigration is set to false
-        await StopEdgeEvents();
+        StopEdgeEvents();
```
lgarner, line 551: This should be handled by the app, since it need to transfer state under its control.

Runtime/Scripts/PersistentConnection.cs
ah1053: No, I can't. I need everything to be called from the UI thread. Will change the boolean to enum as mentioned above.

Runtime/Scripts/PersistentConnection.cs
```diff
@@ -296,104 +355,175 @@ IEnumerator OnIntervalEdgeEventsLatency(EdgeEventsConnection connection, string
...
+        case ServerEventType.EventLatencyRequest:
+          if (fcTriggers.Contains(FindCloudletEventTrigger.LatencyTooHigh))
+          {
+            StartCoroutine(RespondToLatencyRequest());
+          }
+          return;
+        case ServerEventType.EventLatencyProcessed:
+          if (fcTriggers.Contains(FindCloudletEventTrigger.LatencyTooHigh))
+          {
+            if(processingStatus == LatencyProcessingStatus.Ready)
```
franklin-huang-mobiledgex, line 383: nit space between "if" and "("

Runtime/Scripts/PersistentConnection.cs
```diff
@@ -296,104 +355,175 @@ IEnumerator OnIntervalEdgeEventsLatency(EdgeEventsConnection connection, string
...
+            PropagateSuccess(FindCloudletEventTrigger.AppInstHealthChanged, edgeEvent.NewCloudlet);
+          }
+          return;
+        case ServerEventType.EventCloudletMaintenance:
+          if (fcTriggers.Contains(FindCloudletEventTrigger.CloudletMaintenanceStateChanged))
+          {
+            if (edgeEvent.NewCloudlet == null || edgeEvent.ErrorMsg != "")
             {
-              PropagateSuccess(trigger, edgeEvent.NewCloudlet);
+              PropagateError(FindCloudletEventTrigger.CloudletMaintenanceStateChanged, "Received CloudletStateChanged Event but NewCloudlet is null" + edgeEvent.ErrorMsg);
```
franklin-huang-mobiledgex, line 410: nit: add a space after 'null', so that the errorMsg isn't concatenated directly to it

Runtime/Scripts/PersistentConnection.cs
franklin-huang-mobiledgex: What is hostOverride? This really should be the same dme host that was used to start the connection

Runtime/Scripts/PersistentConnection.cs
franklin-huang-mobiledgex: we should return after propagating errors. (eg. we shouldn't be doing FindCloudlet if RegisterClient already failed)

Runtime/Scripts/PersistentConnection.cs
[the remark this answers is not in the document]
ah1053: renamed the boolean to stopEdgeEventsRequested and I assign its value to False at StartEdgeEvents()

Runtime/Scripts/PersistentConnection.cs
ah1053: hostOverride, portOverride are sat in the Initial FindCloudlet.

Runtime/Scripts/PersistentConnection.cs
ah1053: fixed

franklin-huang-mobiledgex (approved):

Runtime/Scripts/PersistentConnection.cs
ah1053: Now logging that Migration occurs and asking developer to StopEdgeEvents on their own.

Runtime/Scripts/PersistentConnection.cs
ah1053: As discussed.

Runtime/Scripts/PersistentConnection.cs
```diff
@@ -296,104 +327,176 @@ IEnumerator OnIntervalEdgeEventsLatency(EdgeEventsConnection connection, string
...
       else
       {
         Logger.Log("Sending Latency Samples (UDP) as a response to LatencyRequest from the server");
-        await integration.matchingEngine.EdgeEventsConnection.TestPingAndPostLatencyUpdate(host, location);
+        integration.matchingEngine.EdgeEventsConnection.TestPingAndPostLatencyUpdate(host, location).ConfigureAwait(false);
       }
     }
 
-    async void ProcessLatency(Statistics stats)
+    async Task ProcessLatency(Statistics stats)
```
lgarner, line 456: Mini Nit: Maybe a tiny Util wrapper that converts Task to IEnumerator for Unity users (while not complete, yield break).

ah1053: Why?!
This is not intended for Unity devs to use. It's a private function.

lgarner (approved):
👍

brucearmstrong (approved):
👍

PR #168, opened 2021-08-09
EDGECLOUD-5456: SignalStrength can be null if no SIM.

lgarner:
On the tin.

### DIFF
--- a/Runtime/Scripts/CarrierInfoIntegration.cs
+++ b/Runtime/Scripts/CarrierInfoIntegration.cs
@@ -178,6 +178,10 @@ namespace MobiledgeX
         return -1;
       }
       AndroidJavaObject signalStrength = telManager.Call<AndroidJavaObject>("getSignalStrength");
+      if (signalStrength == null)
+      {
+        return -1;
+      }
       int signalStrengthLevel =  signalStrength.Call<int>("getLevel");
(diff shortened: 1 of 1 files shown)


franklin-huang-mobiledgex (approved):

brucearmstrong (approved):

Runtime/Scripts/CarrierInfoIntegration.cs
```diff
@@ -178,6 +178,10 @@ public int GetSignalStrength()
         return -1;
       }
       AndroidJavaObject signalStrength = telManager.Call<AndroidJavaObject>("getSignalStrength");
+      if (signalStrength == null)
```
ah1053, line 181: Shall you check for null or INVALID as mentioned in  Android Docs
https://developer.android.com/reference/android/telephony/SignalStrength
Screen Shot 2021-08-24 at 10 33 17 AM

lgarner: SignalStrength's documentedabstracted public getLevel() range is [0-4] inclusive. No SIM or TelManager --> -1. Next to no signal? 0. I guess we happily return INVALID as well depending on getLevel being documented correctly or not, so at least it's distinguishable in the backend. We'll need to see more platforms.

PR #177, opened 2021-09-03
use using() instead of dispose

ah1053:


lgarner (approved):
Was there an issue with putting mxi instantiation inside the using (var mxi = new MobiledgeXIntegration(..) ) ?

Tests/Runtime/MobiledgeX_RuntimeTests.cs
```diff
@@ -52,33 +52,37 @@ public void MobiledgeXEnvironmentSetup()
         public void RegisterClient(string orgName, string appName, string appVers)
         {
             MobiledgeXIntegration mxi = new MobiledgeXIntegration(new CarrierInfoClass(), null, new UniqueIDClass(), new TestDeviceInfo());
-            mxi.appName = appName;
-            mxi.appVers = appVers;
-            mxi.orgName = orgName;
-            var task = Task.Run(async () =>
+            using (mxi)
```
lgarner, line 55: Inner and outer mxi's?

ah1053: Fixed

PR #178, opened 2021-09-04
Reset processingStatus if current latency is less than latency threshold

ah1053:


Runtime/Scripts/EdgeEventsManager.cs
```diff
@@ -616,6 +616,7 @@ void ProcessLatency(Statistics stats)
       else
       {
         Logger.Log("Current Latency (" + stats.Avg + ") < LatencyThreshold (" + config.latencyThresholdTriggerMs + ")");
+        processingStatus = LatencyProcessingStatus.Ready;
```
lgarner, line 619: Ok. I assume nothing is waiting on this status, just for read only whatever wants to start the next test.

lgarner (approved):
LGTM.

PR #181, opened 2021-09-15
Don't stop edge events on error, more extensive logs

ah1053:
1. Don't Stop EdgeEvents when an error occurs.
2. More extensive logs*.

*(JSONUtility.toJSON needs Serializable attribute to be added to C# SDK edge events) and it doesn't support nested deserialization.

### DIFF
--- a/Runtime/Scripts/EdgeEventsManager.cs
+++ b/Runtime/Scripts/EdgeEventsManager.cs
@@ -219,7 +219,21 @@ namespace MobiledgeX
       }
       integration.matchingEngine.EdgeEventsConnection.Open();
       config = MobiledgeXIntegration.settings.edgeEventsConfig;
-      Logger.LogObject(message: "configSummary", obj: config);
+      string configSummary = "EdgeEvents Config Summary:";
+      configSummary += "\nLatency Test Port: " + config.latencyTestPort;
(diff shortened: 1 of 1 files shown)


brucearmstrong:
I don't see a change involving stopping Edge Events. More to come?

Runtime/Scripts/EdgeEventsManager.cs
```diff
@@ -553,7 +573,6 @@ void PropagateError(FindCloudletEventTrigger trigger, string error_msg)
       findCloudletEvent.trigger = trigger;
       Debug.LogError(error_msg);
       EdgeEventsStatus eventStatus = new EdgeEventsStatus(Status.error, error_msg);
-      StopEdgeEvents();
```
ah1053, line 556: Only this change here involving EdgeEvents logic @brucearmstrong 
Basically don't stop edge events if there is an error

lgarner (approved):
Ok. Just simplify the ToString() handling, or a separate ticket to fix the verbosity with an override.

brucearmstrong (approved):

ah1053:
@lgarner  created this for the toString override for EdgeEventsConfig and UpdateConfig https://mobiledgex.atlassian.net/browse/EDGECLOUD-5606

PR #182, opened 2021-09-20
EdgeEventsManager fixes

ah1053:
1. PostLocation&Latency Fire and Forget.
On Unity 2020 this causes InvalidOperationException
Screen Shot 2021-09-20 at 4 23 33 PM
2. Log HostOverride and PortOverride used through EdgeEventsManager.
3. Don't delete locationInfo after stopping EdgeEvents (In case it is used after auto migration)

### DIFF
--- a/Runtime/Scripts/EdgeEventsManager.cs
+++ b/Runtime/Scripts/EdgeEventsManager.cs
@@ -273,7 +273,7 @@ namespace MobiledgeX
     IEnumerator StartEdgeEventsLocation(EdgeEventsConnection connection)
     {
       yield return StartCoroutine(UpdateLocation());
-      connection.PostLocationUpdate(location).ConfigureAwait(false).GetAwaiter().GetResult();
+      connection.PostLocationUpdate(location).ConfigureAwait(false);
       switch (config.locationConfig.updatePattern)
       {
         case UpdatePattern.OnStart:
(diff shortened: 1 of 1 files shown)


Runtime/Scripts/EdgeEventsManager.cs
```diff
@@ -273,7 +273,7 @@ public void StartEdgeEvents(MobiledgeXIntegration mxi)
     IEnumerator StartEdgeEventsLocation(EdgeEventsConnection connection)
     {
       yield return StartCoroutine(UpdateLocation());
-      connection.PostLocationUpdate(location).ConfigureAwait(false).GetAwaiter().GetResult();
+      connection.PostLocationUpdate(location).ConfigureAwait(false);
```
lgarner, line 276: Makes sense, yes. This background operation longer conflicts with DeviceInfo being on the UI thread, right?

Runtime/Scripts/EdgeEventsManager.cs
```diff
@@ -294,13 +294,13 @@ IEnumerator StartEdgeEventsLatency(EdgeEventsConnection connection, string host)
       bool requestSent;
       if (hasTCPPorts)
       {
-        requestSent = connection.TestConnectAndPostLatencyUpdate(host, (uint)config.latencyTestPort, location).ConfigureAwait(false).GetAwaiter().GetResult();
-        Logger.Log("TestConnectAndPostLatencyUpdate : " + requestSent);
+        connection.TestConnectAndPostLatencyUpdate(host, (uint)config.latencyTestPort, location).ConfigureAwait(false);
+        Logger.Log("TestConnectAndPostLatencyUpdate : fired " );
```
lgarner, line 298: nit: space.

Runtime/Scripts/EdgeEventsManager.cs
ah1053: Yes

PR #189, opened 2021-10-04
EdgeEvents gRPC Updates

ah1053:
- New Binary 3.0.3
- Added EdgeEvents Errors similar to Android and iOS SDKs (plus the logging), Example scripts Updated.
- Remove OS dependent code to CarrierInfo instead of DeviceInfo
- Allow override of CarrierInfo in Device Info, for non-cellular devices.
- AutoMigration updated with the new RestartEdgeEventsConnection from the C# binary.
-  Added MEL_ENABLED to MobiledgeXSettings to avoid MEL initialization if MEL is not supported.
- Added ExampleNonCellular for non-cellular devices.
- Added Connection Details class for handling HostOverride, PostOverride.


Runtime/Scripts/ExampleNonCellular.cs
```diff
@@ -0,0 +1,76 @@
...
+    if (fcEvent.newCloudlet != null)
+    {
+      print("New Cloudlet FQDN: " + fcEvent.newCloudlet.Fqdn);
+    }
+    if (edgeEventstatus.status == Status.error)
+    {
+      print("Error received: " + edgeEventstatus.error);
+    }
+  }
+
```
brucearmstrong, line 69: Nit: Extra line

brucearmstrong (approved):

Runtime/Scripts/CarrierInfoIntegration.cs
```diff
@@ -170,19 +193,19 @@ public AndroidJavaObject GetTelephonyManager()
       return telManager;
     }
 
-    public int GetSignalStrength()
+    public ulong GetSignalStrength()
```
lgarner, line 196: This is primarily an Android thing, so it should match what the Android SDK is returning. (0-4, which -1 being not able to return anything, unless 5 has a meaning. There are no unsigned integers in Java.

lgarner: Related API doc: https://developer.android.com/reference/android/telephony/SignalStrength#getLevel()

ah1053: Created a bug to track that https://mobiledgex.atlassian.net/browse/EDGECLOUD-5743

PR #192, opened 2021-10-14
use IDFV instead of IDFA(GRPC)

ah1053:
Same change for REST but this for the gRPC.
https://github.com/mobiledgex/edge-cloud-sdk-unity/pull/191


lgarner (approved):
👍

PR #198, opened 2021-11-23
Location Updates

ah1053:
1. Added a new parameter to LocationService.EnsureLocation Coroutine **obtainLocationOnce** stops location updates if set to true (default is true).
2. Added multiple methods to LocationService (UpdateEdgeEventsManagerLocation, UpdateEdgeEventsLocation, StopLocationUpdates).
3. Call UpdateEdgeEventsManagerLocation from StartEdgeEvents only if it's called from the UI thread.
4. Added IsMainThread property.

### DIFF
--- a/Runtime/Scripts/EdgeEventsManager.cs
+++ b/Runtime/Scripts/EdgeEventsManager.cs
@@ -233,10 +233,6 @@ namespace MobiledgeX
     {
       Logger.Log("Starting EdgeEvents");
       config = MobiledgeXIntegration.settings.edgeEventsConfig;
-      if (IsMainThread) // Input.location can be accessed only from UI Thread
-      {
-        StartCoroutine(locationService.UpdateEdgeEventsManagerLocation(Math.Min(config.locationConfig.updateIntervalSeconds, config.latencyConfig.updateIntervalSeconds)));
-      }
       if (config == null)
       {
         throw new Exception("MobiledgeX EdgeEventsConfig is Null");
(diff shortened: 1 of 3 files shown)


Runtime/Scripts/LocationService.cs
```diff
@@ -84,8 +88,12 @@ public static IEnumerator EnsureLocation()
         {
           if (Input.location.status == LocationServiceStatus.Running)
           {
-            Logger.Log("Location Service succeded, Stopping LocationService, data saved to Input.location.lastData");
-            Input.location.Stop();
+            Logger.Log("Location Service succeded");
+            if (obtainLocationOnce)
```
lgarner, line 92: ok.

Runtime/Scripts/EdgeEventsManager.cs
```diff
@@ -197,6 +233,14 @@ public void StartEdgeEvents(ConnectionDetails connectionDetails)
     {
       Logger.Log("Starting EdgeEvents");
       config = MobiledgeXIntegration.settings.edgeEventsConfig;
+      if (IsMainThread) // Input.location can be accessed only from UI Thread
+      {
+        StartCoroutine(locationService.UpdateEdgeEventsManagerLocation(Math.Min(config.locationConfig.updateIntervalSeconds, config.latencyConfig.updateIntervalSeconds)));
```
lgarner, line 238: What happens if Min == 0?

ah1053: Good point will remove this code block to right after the ConfigValidation so if either of them == 0 it would throw an exception.

ah1053: Fixed, Now if locationConfig.updateIntervalSeconds =0 or latencyConfig.updateIntervalSeconds =0, An error will be produced and EdgeEvents won't start.

brucearmstrong (approved):

PR #200, opened 2022-01-18
EDGECLOUD-4791

ah1053:
Fix for IP To GeoLocation occasionally returns 0,0
https://mobiledgex.atlassian.net/browse/EDGECLOUD-4791

### DIFF
--- a/Runtime/Scripts/MobiledgeXIntegrationConfig.cs
+++ b/Runtime/Scripts/MobiledgeXIntegrationConfig.cs
@@ -90,7 +90,7 @@ namespace MobiledgeX
         LocationFromIPAddress location = Messaging<LocationFromIPAddress>.Deserialize(responseBodyStr);
         if (location.latitude == 0 && location.longitude == 0)
         {
-          throw new Exception("Location API returned {0,0}, overriding location");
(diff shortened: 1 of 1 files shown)


Runtime/Scripts/MobiledgeXIntegrationConfig.cs
```diff
@@ -88,6 +88,10 @@ public static async Task<LocationFromIPAddress> GetLocationFromIP()
         HttpResponseMessage response = await httpClient.GetAsync("https://freegeoip.app/json/").ConfigureAwait(false);
         string responseBodyStr = response.Content.ReadAsStringAsync().Result;
         LocationFromIPAddress location = Messaging<LocationFromIPAddress>.Deserialize(responseBodyStr);
+        if (location.latitude == 0 && location.longitude == 0)
+        {
+          throw new Exception("Location API returned {0,0}, overriding location");
```
brucearmstrong, line 93: It says overriding, but what is the value used for the override?

ah1053: Thanks.
Fixed.

